PR #19, opened 2023-01-09
Add membership event for users not yet lazy loaded

S7evinK:
If a users wasn't lazy loaded, but sends a typing notification, clients don't have a way to display the correct event.
This, in theory, should lazy load members sending typing notifications if the client doesn't know about them yet.


sync3/handler/connstate_live.go
```diff
@@ -197,6 +198,22 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 			if delta.JoinCountChanged {
 				thisRoom.JoinedCount = roomUpdate.GlobalRoomMetadata().JoinCount
 			}
+
+			if s.lazyCache.IsLazyLoading(roomUpdate.RoomID()) && roomUpdate.GlobalRoomMetadata().TypingEvent != nil {
```
kegsay, line 202: This is being done in the wrong place. If you get a typing notification, it appears in the typing extension, whereas this new code only executes if there is an entry in `rooms` (which means there was some kind of event in said room). You need to:
 - Only do this if the typing extension is enabled.
 - Create an entry in `rooms` if one doesn't exist already, for inserting the room member.

kegsay (changes requested):
Wrong place but the code seems good.

kegsay (changes requested):
Looks better, though I dislike how we handle extensions in `connstate_live.go` but the data you need lives there currently. I'll rejig this later.

Requesting that you add an E2E test for this. Something like:
 - Alice make room
 - Bob joins room
 - Alice sends 2 msgs
 - Alice does SS with `timeline_limit: 1` and LL members and typing enabled. We should only get back our own Alice membership.
 - Bob starts typing.
 - Alice should get both typing event and bob's membership event.

tests-e2e/typing_test.go
```diff
@@ -132,6 +132,80 @@ func TestTypingNoUpdate(t *testing.T) {
...
+	// There should only be Alice lazy loaded
+	m.MatchResponse(t, syncResp, m.MatchRoomSubscriptionsStrict(map[string][]m.RoomMatcher{
+		roomID: {
+			MatchRoomRequiredState([]Event{{Type: "m.room.member", StateKey: &alice.UserID}}),
+		},
+	}))
+
+	// Bob starts typing
+	bob.SendTyping(t, roomID, true, 5000)
+
```
kegsay, line 184: Everything above this comment is great. Everything below can be optimised into 2 lines. When Bob sends typing, we don't know when the proxy will get it. It is not safe to do a single sliding sync call like you do below.

Something like:
```go
syncResp = waitUntilTypingData(t, alice, roomID, []string{bob.UserID})
m.MatchResponse(t, syncResp, m.MatchRoomSubscriptionsStrict(map[string][]m.RoomMatcher{
    roomID: {
        MatchRoomRequiredState([]Event{{Type: "m.room.member", StateKey: &bob.UserID}}),
    }
}))

kegsay (approved):

PR #20, opened 2023-01-12
Add docker GHA workflow

S7evinK:
This adds a docker GHA workflow which builds a `latest` tag on every push to `main` and a release version on releases. (untested, taken from Dendrite)
The dockerfile is slightly updated to make use of buildkit, allowing for faster builds.
Some dependencies were updated, since they contained vulns and would be reported by Trviy otherwise.

Unsure about the defined platforms or if we just want/need `linux/amd64`


.github/workflows/tests.yml
```diff
@@ -109,7 +109,9 @@ jobs:
         steps:
             - uses: actions/checkout@v2
             - name: "Build docker image"
-              run: docker build -f Dockerfile -t ghcr.io/matrix-org/sliding-sync-proxy:ci  .
+              env:
+                DOCKER_BUILDKIT: 1
+              run: docker build -f Dockerfile --target=sliding-sync -t ghcr.io/matrix-org/sliding-sync-proxy:ci  .
```
kegsay, line 114: Why specify a `--target` here? `sliding-sync` is the last `FROM` already, this just adds noise.

Dockerfile
```diff
@@ -1,17 +1,19 @@
...
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg/mod \
+  GIT_COMMIT=$(git rev-list -1 HEAD) && \
+  go build -ldflags "-X main.GitCommit=$GIT_COMMIT" -trimpath -o /out/ ./cmd/syncv3
 
-FROM alpine:latest
-
-COPY --from=base /build/bin/* /usr/bin/
+FROM alpine:3.17 AS sliding-sync
```
kegsay, line 12: Pointless `AS` imo.

kegsay (approved):
Otherwise LGTM! Thanks!

PR #30, opened 2023-02-27
Add YouTube link to Landing.md

pixlwave:


kegsay:
The youtube video will go out-of-date at some point in the future, at which point we would now need to remember to update this or remove this. I'd rather we didn't.

pixlwave:
Fair point :)

PR #32, opened 2023-03-15
WIP: Only reorder the room lists on seeing specific event types

DMRobertson:


internal/roomname.go
```diff
@@ -14,7 +14,7 @@ type RoomMetadata struct {
 	CanonicalAlias       string
 	JoinCount            int
 	InviteCount          int
-	LastMessageTimestamp uint64
+	LastMessageTimestamp uint64 // TODO: remove this?
```
kegsay, line 17: No because we need this as a fallback.

sync3/sort_test.go
```diff
@@ -80,6 +83,7 @@ func TestSortBySingleOperation(t *testing.T) {
 				NotificationCount: 1,
 				CanonicalisedName: "boo",
 			},
+			LastActivityTimestamp: 800,
```
kegsay, line 86: It's very unfortunate we need to duplicate the timestamp here..

tests-e2e/lists_test.go
```diff
@@ -856,3 +856,121 @@ func TestMultipleSameList(t *testing.T) {
...
+		},
+	}
+	aliceSyncRequest := sync3.Request{
+		Lists: map[string]sync3.RequestList{
+			"room_list": aliceReqList,
+		},
+		BumpEventTypes: []string{"m.room.message", "m.room.encrypted"},
+	}
+	// TODO: to avoid flakes, want this sync to include/cover Bob's messages in both rooms.
+	aliceRes := alice.SlidingSync(t, aliceSyncRequest)
```
kegsay, line 918: It will because of how sliding sync works. It's lazy, so the first time you call sliding sync it will do an initial sync, which will return the most recent event in each room. This will be the two hello world messages, so there's no race conditions here, but only because the first SS request blocks on upstream /sync.

tests-e2e/lists_test.go
```diff
@@ -856,3 +856,121 @@ func TestMultipleSameList(t *testing.T) {
...
+	matchNoOps := m.MatchList("room_list",
+		m.MatchV3Count(2),
+		m.MatchV3Ops(),
+	)
+	m.MatchResponse(t, aliceRes, matchNoOps)
+
+	t.Log("Bob syncs until he sees Charlie's membership.")
+	bobRes = bob.SlidingSyncUntilMembership(t, bobRes.Pos, room2, charlie, "join")
+
+	t.Log("Bob should see room 2 at the top of his list.")
```
kegsay, line 964: These logs are unhelpful when things fail because I at least ask "why?" and that isn't explained in the logs. Why should Bob see room 2 at the top of his list? Also, the dev can probably ascertain what is being asserted by just reading the code.

tests-e2e/lists_test.go
```diff
@@ -856,3 +856,121 @@ func TestMultipleSameList(t *testing.T) {
...
+	t.Log("Charlie joins room 2.")
+	charlie.JoinRoom(t, room2, nil)
+
+	t.Log("Alice syncs until she sees Charlie's membership.")
+	aliceRes = alice.SlidingSyncUntilMembership(t, aliceRes.Pos, room2, charlie, "join")
+
+	t.Log("Alice shouldn't see any rooms' positions change.")
+	matchNoOps := m.MatchList("room_list",
+		m.MatchV3Count(2),
+		m.MatchV3Ops(),
```
kegsay, line 957: `m.MatchNoV3Ops()` would be better.

sync3/handler/connstate_live.go
```diff
@@ -284,22 +284,45 @@ func (s *connStateLive) processUpdatesForSubscriptions(ctx context.Context, buil
...
+		} else {
+			roomActivityTimestamp = roomEventUpdate.EventData.Timestamp
+		}
 	}
+
+	inviteEventUpdate, ok := up.(*caches.InviteUpdate)
+	if ok {
+		// Even if we have ignored membership events, we probably want to know when
+		// we've been invited to a room.
+		roomActivityTimestamp = inviteEventUpdate.InviteData.LastMessageTimestamp
```
kegsay, line 313: I feel uneasy with this because it feels like we're going to inevitably add in a lot of exceptions. I wonder if a better approach would be to just modify the `LastMessageTimestamp` directly, instead of sometimes copying over the value, sometimes sending 0 to mean no change, etc. This would also mean the tests don't need to duplicate the timestamp, and should make it much less likely to break everything since you're just very situationally (if `bump_event_types` is set) modifying the `LastMessageTimestamp` rather than fundamentally changing the sort logic to use `LastActivityTimestamp`.

We reliably return copies of `RoomMetadata` so you can edit it freely and not pollute other connections, see https://github.com/matrix-org/sliding-sync/blob/main/sync3/caches/global.go#L79

We have precedent for modifying this struct when calculating the room name, as we modify the `Heroes` slice to remove the user ID of the caller, see https://github.com/matrix-org/sliding-sync/blob/main/sync3/handler/connstate.go#L94 and https://github.com/matrix-org/sliding-sync/blob/main/internal/roomname.go#L65

kegsay (changes requested):
Mostly there, and with using `LastMessageTimestamp` more you should find your test failures disappear.

tests-e2e/lists_test.go
DMRobertson: I take the point that this doesn't spell out "why", but I think that can be inferred from looking at the full sequence of these signpost logs. (Which doesn't seem all that different to reading the source, to me.)

> These logs are unhelpful when things fail

Strong disagree here. We have found this pattern useful in complement, particularly in longer tests with lots of output. Pain points:

- not easy to see which clients are making CSAPI requests (e.g. who is syncing?)
- easy to miss a `t.Errorf` that happened midway through the test
- Errorf/Fatalf text isn't red or otherwise made more prominent (e.g. by being printed separately to the test logs)

sync3/handler/connstate_live.go
DMRobertson: Moreover it seems like that heroes modification ought to be done in `UserRoomData` rather than `RoomMetadata`?

DMRobertson:
Superseded by #34.

Issue #41, opened 2023-03-27
Expand extensions to allow them to be registered on a specific named list(s)


kegsay:
This is so background spider lists don’t get to-device messages / typing / receipts / account data. - We can do this now because key-based lists are landed. This is useful for EX because the “all rooms” list we don’t want typing notifs / receipts / etc, but we do in the “visible rooms” list, and atm you can’t express that.

SITREP:

- [x] typing #50
- [x] account data #53
- [x] receipts #54
- [ ] todevice
- [x] e2ee: N/A. None of the information communicated by this extension is scoped to a particular room.

DMRobertson:
Are you envisioning this as something that each extension opts into, or a mechanism baked into the MSC (e.g. a mandatory `lists` key reserved by the spec within each extension's parameter object)?

> This is useful for EX because the “all rooms” list we don’t want typing notifs / receipts / etc, but we do in the “visible rooms” list, and atm you can’t express that.

Do we have a similar situation with the Room Subscriptions API? Is it useful for EX to subscript to a room but e.g. opt out of typing notifications for that room?

kegsay:
> (e.g. a mandatory lists key reserved by the spec within each extension's parameter object)?

Yes, this. It would be useful to have a single way to express "this extension applies to these named lists and these named room IDs". This single way could just be a naming convention e.g every extension can say `lists: [ "foo" ,"bar" ]` and `rooms: ["!foo", "!bar"]` .

> Do we have a similar situation with the Room Subscriptions API? Is it useful for EX to subscript to a room but e.g. opt out of typing notifications for that room?

Potentially, yes.

I'm unsure on the best name for the key. In essence, the values represent the overall scope of that extension, and its omission will catch all lists/rooms on the user's account. It's important not to imply it's returning room events because in some cases it might not be (e.g the to-device extension doesn't have room events so you might think this is pointless, but it's actually useful in the future if we want to say only deliver to-device events for users who are in these rooms given by these lists. The values can be either a room ID or a list key. It's _probably_ safe to mux them together given list keys shouldn't start with a `!`, and you're only confusing yourself if you name a list after a room ID.  Maybe it's good enough to just have it be: `scope: [ "list_a", "list_b", "!roomA, "!roomB" ]` ?

DMRobertson:
> I'll write something down on the MSC and see how it goes.

[...]

PR #53, opened 2023-03-31
List/room granularity for the account data extension

DMRobertson:
More of #41.

This was easier than I expected; the hardest part was writing the test.

I'd intended for one of the tests I wrote in #50 to demonstrate that I had correctly changed `AppendLive`, but I think the test I wrote simply doesn't do that.

I think what I've got here really does exercise the `AppendLive` path---but it does so as an e2e test and is probably going to be brittle (e.g. there are timeouts). Maybe this would be better suited for`extensions/account_data_test.go`?

### DIFF
--- a/tests-e2e/account_data_test.go
+++ b/tests-e2e/account_data_test.go
@@ -7,7 +7,6 @@ import (
 	"github.com/matrix-org/sliding-sync/testutils"
 	"github.com/matrix-org/sliding-sync/testutils/m"
 	"testing"
-	"time"
 )
 
 func TestAccountDataRespectsExtensionScope(t *testing.T) {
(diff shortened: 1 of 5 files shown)


tests-e2e/account_data_test.go
```diff
@@ -0,0 +1,160 @@
...
+					},
+				},
+			},
+			WithPos(pos),
+		)
+		pos = response.Pos
+		responses <- response
+	}
+
+	go aliceIncrementalSync()
```
kegsay, line 99: Please don't use goroutines unless you absolutely have to. You don't have to here. It's sufficient to just call `putGlobalAccountData` and then synchronously call `alice.SlidingSyncUntil(...)` to await for the response. This handles timeouts for you.

kegsay (changes requested):
Tests need to be simpler and not have channels/goroutines, otherwise this is excellent, LGTM

tests-e2e/account_data_test.go
DMRobertson: The thinking here: I was trying to test that `AppendLive` has the correct logic in place. If the proxy is already aware of a change before the sync request comes in, that change gets picked up by `ProcessInitial` and a response is returned immediately, without hitting `AppendLive`.

DMRobertson: This thinking was not correct. See 391f0b8

kegsay (approved):

Issue #21, opened 2023-01-23
Add Sentry support


kegsay:
To make certain errors much more visible.

DMRobertson:
First step would be something that reports errors in responding to incoming sync v3 requests.

DMRobertson:
https://github.com/archdx/zerolog-sentry might be nice to use

DMRobertson:
I think #57 and #67 are a good first pass. It'd be good to feed some more information into sentry (e.g. [usernames](https://docs.sentry.io/platforms/go/enriching-events/identify-user/) or the other tricks described at https://docs.sentry.io/platforms/go/enriching-events/). But it's probably best to see how we use the Sentry reports in practice before going further.

DMRobertson:
Other things I'd like to do:

- [associate incoming requests to a user](https://docs.sentry.io/platforms/go/enriching-events/identify-user/)
- add logging for panics that won't get `recover`ed from

PR #57, opened 2023-04-04
Rough first pass of sentry support

DMRobertson:
Part of #21.

Cribbed this from https://docs.sentry.io/platforms/go/guides/http/ and dendrite.

This should be enough to make internal server errors and panics on v3 request handling get reported to sentry.

In an upcoming PR I will match up every error log line with a sentry `CaptureException` call.


DMRobertson:
I was able to test this manually with a call to `sentry.CaptureMessage("HELLO SENTRY!")`.

image

cmd/syncv3/main.go
```diff
@@ -113,10 +117,31 @@ func main() {
...
+		}
+	}
+
 	go h2.StartV2Pollers()
 	if args[EnvJaeger] != "" {
 		h3 = otelhttp.NewHandler(h3, "Sync")
 	}
+
+	if args[EnvSentryDsn] != "" {
+		defer sentry.Flush(2 * time.Second)
```
kegsay, line 138: This will never execute because `defer` runs at the end of the function, and this function will never end due to `select {} // block forever`

DMRobertson: Apparently not. I can crib from Dendrite, e.g. https://github.com/matrix-org/dendrite/blob/236b16aa6c97bc0894388dce7f6b420ef7a1fd88/setup/base/base.go#L524

cmd/syncv3/main.go
```diff
@@ -113,10 +117,31 @@ func main() {
...
+		sentryHandler := sentryhttp.New(sentryhttp.Options{
+			Repanic: true,
+		})
+		h3 = sentryHandler.Handle(h3)
```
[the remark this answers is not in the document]
DMRobertson: I think [faef68b](https://github.com/matrix-org/sliding-sync/pull/57/commits/faef68bc6fee1cdd13ab75913ea978a253b78197) might be more what you have in mind(?)

cmd/syncv3/main.go
DMRobertson: https://github.com/matrix-org/sliding-sync/pull/57/commits/40d9f76b9b64c4a641b85c72c8b534039a19acd0 blocks ditches the infinite select for a SIGINT/TERM

sync3/conn.go
```diff
@@ -83,6 +84,7 @@ func (c *Conn) tryRequest(ctx context.Context, req *Request) (res *Response, err
 		if panicErr != nil {
 			err = fmt.Errorf("panic: %s", panicErr)
 			logger.Error().Msg(string(debug.Stack()))
+			sentry.GetHubFromContext(ctx).RecoverWithContext(ctx, panicErr)
```
kegsay, line 87: There won't ever be a hub though? I guess we do need the sentry middleware if we want to log contextual information about the request. Bleh.

cmd/syncv3/main.go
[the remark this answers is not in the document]
DMRobertson: I suppose we could copy bits of https://github.com/getsentry/sentry-go/blob/master/http/sentryhttp.go#L84-L106 to make hubs for ourselves, but I'm more inclined to let Sentry do it for us!

cmd/syncv3/main.go
```diff
@@ -113,10 +120,53 @@ func main() {
 		AddPrometheusMetrics: args[EnvPrometheus] != "",
 	})
 
+	if args[EnvSentryDsn] != "" {
+		fmt.Printf("Configuring Sentry reporter...\n")
+		err := sentry.Init(sentry.ClientOptions{
```
kegsay, line 125: Why isn't this with the other `if args[EnvSentryDsn] != "" {` code? I'm guessing we need to have init'd it prior to calling `go h2.StartV2Pollers()` - but why? Please comment.

sync3/conn.go
```diff
@@ -78,11 +78,15 @@ func (c *Conn) OnUpdate(ctx context.Context, update caches.Update) {
 }
 
 func (c *Conn) tryRequest(ctx context.Context, req *Request) (res *Response, err error) {
+	// TODO: include useful information from the request in the sentry hub/context
 	defer func() {
 		panicErr := recover()
 		if panicErr != nil {
 			err = fmt.Errorf("panic: %s", panicErr)
 			logger.Error().Msg(string(debug.Stack()))
+			internal.GetSentryHubFromContextOrDefault(ctx).RecoverWithContext(ctx, panicErr)
```
DMRobertson, line 87: This is a bit of a mouthful. Not wedded to the name.

I only need the fallback hub for one test case in this PR. However in an upcoming PR I want to send events to sentry with a context that might have a hub or might not.

cmd/syncv3/main.go
```diff
@@ -113,10 +120,53 @@ func main() {
...
+		}
+	}
+
 	go h2.StartV2Pollers()
 	if args[EnvJaeger] != "" {
 		h3 = otelhttp.NewHandler(h3, "Sync")
 	}
+
+	if args[EnvSentryDsn] != "" {
+		defer sentry.Flush(2 * time.Second)
```
kegsay, line 141: We surely don't need to flush here as we flushed in `WaitForShutdown`?

sync3/conn.go
```diff
@@ -78,11 +78,15 @@ func (c *Conn) OnUpdate(ctx context.Context, update caches.Update) {
...
+		} else if err != nil {
+			internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(err)
```
kegsay, line 89: Why are these calling different things? What is the difference between them? We just want to send up the error, nothing more, so `CaptureException` seems the right thing to do?

kegsay (approved):
Otherwise LGTM

sync3/conn.go
kegsay: Agreed!

cmd/syncv3/main.go
DMRobertson: Err, yes. I think I did a sloppy revert.

cmd/syncv3/main.go
DMRobertson: Two concerns:

- in an upcoming PR, I'm going to make any error messages from `h2.StartV2Pollers` also log to sentry, so I want to init before that.
- I think I was worried  that the order of installing the otel and Sentry handlers might be important. On reflection I'd hope that the two middlewares are orthogonal.

More generally: Sentry has some support for tracking and tracing; I wonder if there's some kind of overlap or duplicated effort here.

DMRobertson: Commented in https://github.com/matrix-org/sliding-sync/pull/57/commits/646232dcb0e6f6daa82dc8dd7770cbf121bac001.

sync3/conn.go
DMRobertson: Commented in https://github.com/matrix-org/sliding-sync/pull/57/commits/646232dcb0e6f6daa82dc8dd7770cbf121bac001.

DMRobertson: But TL;DR: I think sentry might mark `panicErr` as having come from a panic with the former function.

PR #58, opened 2023-04-04
Exact SYNC and INVALIDATE ranges

### LINKED ISSUE #52: Do not echo the `range` from the client
Hi 👋,

Imagine the following scenario:

* The client is asking for a list with a range of `0..=9`, so 10 items,
* The server finds 1 room
* The server responds with a `SYNC` operation with a range of `[0, 9]`.

And this is wrong. It should reply with a range of `[0, 0]` because there is only one room found, not 10.

As discussed online, it's a bug. Right now, the server simply echoes back the range from the client's request, but it should compute a new range instead.

DMRobertson:
Closes #52.

The commit history may be enlightening, but there is a lot of churn.

Each of the three changes I made independently required tests to be fixed up, so I think we can get away without new test coverage. I inspected all four uses of ResponseOpRange.ResponseOpRange and have updated three; the remainder is

https://github.com/matrix-org/sliding-sync/blob/2970cf9f2721a827a3b50ac211ed2bd603105cef/sync3/handler/connstate.go#L200-L218

which never exhibits the kind of overflow we want to avoid.

### DIFF
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -240,10 +240,15 @@ func (s *ConnState) onIncomingListRequest(ctx context.Context, builder *RoomsBui
 		if prevReqList != nil {
 			// there were previous syncs for this list, INVALIDATE the lot
 			logger.Trace().Interface("range", prevRange).Msg("INVALIDATEing because sort/filter ops have changed")
+			allRoomIDs := roomList.RoomIDs()
 			for _, r := range prevRange {
+				clamped := clampSliceRangeToListSize(r, len(allRoomIDs))
+				if clamped == nil {
+					continue
+				}
 				responseOperations = append(responseOperations, &sync3.ResponseOpRange{
 					Operation: sync3.OpInvalidate,
-					Range:     r[:],
+					Range:     clamped,
 				})
 			}
 		}
@@ -266,7 +271,9 @@ func (s *ConnState) onIncomingListRequest(ctx context.Context, builder *RoomsBui
 	for i := range removedRanges {
 		responseOperations = append(responseOperations, &sync3.ResponseOpRange{
 			Operation: sync3.OpInvalidate,
-			Range:     removedRanges[i][:],
+			// we shouldn't need to clamp this range: we are removing concrete rooms
+			// whose position is in bounds.
+			Range: removedRanges[i][:],
 		})
 	}
 
@@ -285,9 +292,13 @@ func (s *ConnState) onIncomingListRequest(ctx context.Context, builder *RoomsBui
 		// the builder will populate this with the right room data
 		builder.AddRoomsToSubscription(subID, roomIDs)
 
+		clamped := clampSliceRangeToListSize(addedRanges[i], len(roomIDs))
+		if clamped == nil {
+			continue
+		}
 		responseOperations = append(responseOperations, &sync3.ResponseOpRange{
 			Operation: sync3.OpSync,
-			Range:     clampSliceRangeToRooms(addedRanges[i], roomIDs),
+			Range:     clamped,
 			RoomIDs:   roomIDs,
 		})
 	}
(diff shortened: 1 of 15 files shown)


Hywan:
:+1:

sync3/handler/connstate.go
```diff
@@ -543,3 +556,25 @@ func (s *ConnState) OnRoomUpdate(ctx context.Context, up caches.RoomUpdate) {
 		logger.Warn().Str("room_id", up.RoomID()).Msg("OnRoomUpdate unknown update type")
 	}
 }
+
+// clampSliceRangeToListSize helps us send client-friend SYNC and INVALIDATE ranges.
```
kegsay, line 560:
```suggestion
// clampSliceRangeToListSize helps us to send client-friendly SYNC and INVALIDATE ranges.
```

sync3/handler/connstate.go
```diff
@@ -543,3 +556,25 @@ func (s *ConnState) OnRoomUpdate(ctx context.Context, up caches.RoomUpdate) {
...
+//
+// Suppose the client asks for a window on positions [10, 19]. If the list
+// has exactly 12 rooms, the window will see 3: those at positions 10, 11
+// and 12. In this situation, it is helpful for clients to be given the
+// "effective" range [10, 12] rather than the "conceptual" range [10, 19].
+//
+// The "full" room list occupies positions [0, totalRooms - 1]. If the given range r
+// does not overlap the full room list, return nil. Otherwise, return the intersection
+// of r with the full room list.
+func clampSliceRangeToListSize(r [2]int64, totalRooms int64) []int64 {
```
kegsay, line 570: This signature could be nicer if we just pass through `r` unaltered in cases where it doesn't reach / is beyond `totalRooms` and then clamp if the range is within the index position. That way, rather than callers being forced to `if nil` check, they can just in-line it e.g:
```
Range: clampSliceRangeToListSize(removedRanges[i][:], totalRooms)
```

I do not expect this function to perform validation on the range, as that should have already happened.

tests-e2e/lists_test.go
```diff
@@ -726,8 +726,8 @@ func TestChangeSortOrder(t *testing.T) {
 	}
 
 	m.MatchResponse(t, res, m.MatchList("a", m.MatchV3Count(4), m.MatchV3Ops(
-		m.MatchV3InvalidateOp(0, 20),
-		m.MatchV3SyncOp(0, 20, []string{gotNameToIDs["Apple"], gotNameToIDs["Kiwi"], gotNameToIDs["Lemon"], gotNameToIDs["Orange"]}),
+		m.MatchV3InvalidateOp(0, 3),
```
kegsay, line 729: Nice, so INVALIDATE _was_ affected.

kegsay (approved):
Excellent, thank you.

sync3/handler/connstate.go
DMRobertson: > This signature could be nicer if we just pass through r unaltered in cases where it doesn't reach / is beyond totalRooms and then clamp if the range is within the index position.

I can see that this works where the range doesn't reach the totalRooms (line 577). But when the start of the range is already beyond total rooms (572), I don't want to tell the client about this range at all---hence the nil return value and `continue`s. But perhaps by

> I do not expect this function to perform validation on the range, as that should have already happened.

you mean that the start of the range `r[0]` should never be beyond `totalRooms`? If so I don't follow, since the start of the range comes from the client.

DMRobertson: Let's write a test to see if this is actually possible in practice.

DMRobertson: The test

```go
// Tests the scenario described at
// https://github.com/matrix-org/sliding-sync/pull/58#discussion_r1159850458
func TestRangesThatExceedTotalRooms(t *testing.T) {
	alice := registerNewUser(t)

	t.Log("Alice makes three public rooms.")
	room0 := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat", "name": "A"})
	room1 := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat", "name": "B"})
	alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat", "name": "C"})

	t.Log("Alice initial syncs, requesting room ranges [0, 1] and [8, 9]")
	syncRes := alice.SlidingSync(t, sync3.Request{
		Lists: map[string]sync3.RequestList{
			"a": {
				Sort:   []string{sync3.SortByName},
				Ranges: sync3.SliceRanges{{0, 1}, {8, 9}},
			},
		},
	})

	t.Log("Alice should only see rooms 0 and 1 in the sync response.")
	m.MatchResponse(
		t,
		syncRes,
		m.MatchList(
			"a",
			m.MatchV3Count(3),
			m.MatchV3Ops(
				m.MatchV3SyncOp(0, 1, []string{room0, room1}),
			),
		),
	)
}
```

passes on main (690790e46d8d477ede3b5a1f21f27726d3fb7a06). So this doesn't seem to happen in practice, presumably due to the continue in

 https://github.com/matrix-org/sliding-sync/blob/main/sync3/handler/connstate.go#L276-L293

Will amend the clamping function as above.

DMRobertson: > Will amend the clamping function as above.

Having done this, we are now seeing assertion failures showing that this logic was necessary, e.g. https://github.com/matrix-org/sliding-sync/actions/runs/4667271913/jobs/8262861981?pr=58#step:7:394

I can either
- go back to returning a possibly nil slice
- fixup the logic so that we never try to clamp an out-of-bounds range.

Any preference?

DMRobertson: > fixup the logic so that we never try to clamp an out-of-bounds range.

Have updated the PR to do this.

kegsay (approved):

PR #61, opened 2023-04-05
Send error logs to Sentry

DMRobertson:
Part of #21. Based on #57.

I looked at the source for all calls to `logger.(Err|Error|Panic|Fatal)` and added an accompanying call to `sentry.CaptureException`. I also did some context plumbing to try tie as many Sentry events to the syncv3 requests that triggered them.

The diff is quite large. I am happy to split this up into smaller PRs if that would help.

### DIFF
--- a/cmd/syncv3/main.go
+++ b/cmd/syncv3/main.go
@@ -115,14 +115,10 @@ func main() {
 			panic(err)
 		}
 	}
-	h2, h3 := syncv3.Setup(args[EnvServer], args[EnvDB], args[EnvSecret], syncv3.Opts{
-		Debug:                args[EnvDebug] == "1",
-		AddPrometheusMetrics: args[EnvPrometheus] != "",
-	})
 
 	// Initialise sentry. We do this in a separate block to the sentry code below,
-	// because we want to configure logging before starting the pollers—they may want to
-	// log to sentry themselves.
+	// because we want to configure logging before the call to syncv3.Setup---which may
+	// want to log to sentry itself.
 	if args[EnvSentryDsn] != "" {
 		fmt.Printf("Configuring Sentry reporter...\n")
 		err := sentry.Init(sentry.ClientOptions{
@@ -136,11 +132,17 @@ func main() {
 		}
 	}
 
+	h2, h3 := syncv3.Setup(args[EnvServer], args[EnvDB], args[EnvSecret], syncv3.Opts{
+		Debug:                args[EnvDebug] == "1",
+		AddPrometheusMetrics: args[EnvPrometheus] != "",
+	})
+
 	go h2.StartV2Pollers()
 	if args[EnvJaeger] != "" {
 		h3 = otelhttp.NewHandler(h3, "Sync")
 	}
 
+	// Install the Sentry middleware, if configured.
 	if args[EnvSentryDsn] != "" {
 		sentryHandler := sentryhttp.New(sentryhttp.Options{
 			Repanic: true,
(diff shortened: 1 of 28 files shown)


DMRobertson:
I should also do an audit for panic() calls and stick in logging. Can make that a separate change though.

cmd/syncv3/main.go
```diff
@@ -115,31 +115,34 @@ func main() {
...
 	if args[EnvSentryDsn] != "" {
 		fmt.Printf("Configuring Sentry reporter...\n")
 		err := sentry.Init(sentry.ClientOptions{
-			Dsn:     args[EnvSentryDsn],
-			Release: version,
-			Dist:    GitCommit,
+			Dsn:         args[EnvSentryDsn],
+			Release:     version,
+			Dist:        GitCommit,
+			Environment: "dmr-test",
```
kegsay, line 128: Probably not intended to be included? :D

sync3/caches/global.go
```diff
@@ -72,22 +73,24 @@ func NewGlobalCache(store *state.Storage) *GlobalCache {
...
 	defer c.roomIDToMetadataMu.RUnlock()
 	result := make(map[string]*internal.RoomMetadata, len(roomIDs))
 	for i := range roomIDs {
 		roomID := roomIDs[i]
 		sr := c.roomIDToMetadata[roomID]
 		if sr == nil {
-			logger.Error().Str("room", roomID).Msg("GlobalCache.LoadRoom: no metadata for this room")
+			const errMsg = "GlobalCache.LoadRoom: no metadata for this room"
+			logger.Error().Str("room", roomID).Msg(errMsg)
+			internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(fmt.Errorf(errMsg))
```
kegsay, line 93: This isn't sentry worthy, it can happen when rooms are initially loaded into the cache.

sync3/caches/user.go
```diff
@@ -138,9 +138,9 @@ func NewInviteData(userID, roomID string, inviteState []json.RawMessage) *Invite
 		}
 	}
 	if id.InviteEvent == nil {
-		logger.Error().Str("invitee", userID).Str("room", roomID).Int("num_invite_state", len(inviteState)).Msg(
-			"cannot make invite, missing invite event for user",
-		)
+		const errMsg = "cannot make invite, missing invite event for user"
+		logger.Error().Str("invitee", userID).Str("room", roomID).Int("num_invite_state", len(inviteState)).Msg(errMsg)
+		internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(fmt.Errorf(errMsg))
```
kegsay, line 143: Can we track the `num_invite_state` etc with this?

kegsay (approved):

cmd/syncv3/main.go
DMRobertson: Oops!

sync3/caches/global.go
DMRobertson: Okay. Shall we downgrade to a warning then?

sync3/caches/user.go
DMRobertson: Looks like a "context" on a "scope" is the way to do this:

E.g. with the test snippet

```go
	hub := internal.GetSentryHubFromContextOrDefault(ctx)
	hub.WithScope(func(scope *sentry.Scope) {
		scope.SetContext("sliding-sync", map[string]interface{}{
			"invitee":          "@dmr:test",
			"room":             "!dmr:test",
			"num_invite_state": 12345,
		})
		hub.CaptureException(fmt.Errorf("DMR TEST DMR TEST"))
	})
```

I get

image

DMRobertson: (You can also write key-value string pairs which get indexed and are searchable but I didn't think that would be useful here.)

PR #63, opened 2023-04-11
bugfix: fix #56 by sending the correct NID with each live event

### LINKED ISSUE #56: Missing messages after a batch of multiple messages are sent in short succession
> @kegsay I can see the problem on EW with SS too. I reproduce the same scenario by switching off and on my network while adding more messages to the room. Some messages are missing too.
>
> Steps to reproduce on EW is similar
> - Have a room 
> - Switch off the network
> - Send enough messages to this room (I sent 0..9 and a..q in my test)
> - Switch on the network
> -> Some messages are missing: 6, 7, 9, B, D, I, J, K:
>
> image

_Originally posted by @manuroe in https://github.com/matrix-org/matrix-rust-sdk/issues/1742#issuecomment-1495683074_

kegsay:
Early versions of the proxy tended to send a list of event JSONs and the "latest" NID for that batch. This then interacted badly with later code which used these NIDs to determine if the event in question should be returned to the client or not. We sometimes filter them out in cases where the "initial" room has already included this event, e.g room has msgs A,B,C which were pulled in initially via a DB call, then we receive C down as an update, we should not include it else we will send back A,B,C,C. By only sending the "latest" NID, we will filter out other events in that batch as they are <= the previously seen latest NID.

This was not tested in E2E tests because it relies on slow pollers which cause >1 timeline event for a single room to arrive. This may be a cause of flakey tests. We now have an integration test for this which injected batches of events for the same room and ensures they are all seen down the connection.

Thanks to @jplatte and @manuroe for helping debug this.

### DIFF
--- a/sync3/dispatcher.go
+++ b/sync3/dispatcher.go
@@ -100,7 +100,9 @@ func (d *Dispatcher) OnNewInitialRoomState(ctx context.Context, roomID string, s
 		logger.Warn().Int("join_count", jc).Str("room", roomID).Int("num_state", len(state)).Msg(
 			"OnNewInitialRoomState but have entries in JoinedRoomsTracker already, this should be impossible. Degrading to live events",
 		)
-		d.OnNewEvents(ctx, roomID, state, 0)
+		for _, s := range state {
+			d.OnNewEvent(ctx, roomID, s, 0)
+		}
 		return
 	}
 	// create event datas for state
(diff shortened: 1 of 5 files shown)


jplatte:
Nice, but can you remove the `@` from my username in the commit message so I don't get pinged whenever the commit moves in some way? 😅

sync3/dispatcher.go
```diff
@@ -141,24 +143,15 @@ func (d *Dispatcher) OnNewInitialRoomState(ctx context.Context, roomID string, s
 	}
 }
 
-// Called by v2 pollers when we receive new events
-func (d *Dispatcher) OnNewEvents(
```
DMRobertson, line 145: Are we just getting ride of this for ergonomics' sake? I.e. was there something buggy about this?

Perhaps the point is that we want to force the caller to thinking about handling events with different `pos`es?

tests-e2e/num_live_test.go
```diff
@@ -122,7 +122,6 @@ func TestNumLive(t *testing.T) {
 	m.MatchResponse(t, res, m.MatchRoomSubscriptionsStrict(map[string][]m.RoomMatcher{
 		roomID: {
 			m.MatchNumLive(2),
-			// TODO: should we be including event ID 2 given timeline limit is 2?
```
DMRobertson, line 125: Don't really follow this---why have we now deeemed this behaviour desirable?

sync3/handler/handler.go
```diff
@@ -512,7 +512,9 @@ func (h *SyncLiveHandler) Accumulate(p *pubsub.V2Accumulate) {
 	}
 	internal.Logf(ctx, "room", fmt.Sprintf("%s: %d events", p.RoomID, len(events)))
 	// we have new events, notify active connections
-	h.Dispatcher.OnNewEvents(ctx, p.RoomID, events, p.EventNIDs[len(p.EventNIDs)-1])
+	for i := range events {
+		h.Dispatcher.OnNewEvent(ctx, p.RoomID, events[i], p.EventNIDs[i])
+	}
```
DMRobertson, line 517: I guess this is the crux of the fix?

sync3/dispatcher.go
kegsay: Yes, it was originally useful for ergonomics to avoid needing callers to for .. range it. However, in practice we would need to for .. range it anyway to link the event NID and the JSON, so we might as well just remove the bulk function.

tests-e2e/num_live_test.go
kegsay: I've removed this mostly because A) it's always been just for the initial limit and B) I originally had the test in this file before realising it needed to be an integration test.

sync3/handler/handler.go
kegsay: Yes

DMRobertson (approved):
I'm happy if CI is happy.

PR #71, opened 2023-04-13
Inject state blocks from incremental polls into the room timeline

DMRobertson:
Closes #18. This is a best-effort attempt to ensure that room state self-heals after the poller "misses" a state event. We don't attempt to retrieve any missed message events.

For a more robust solution, the homeserver should natively implement sliding sync.

#73 also helps, by making it less likely that the poller sees a gap in the first place.

PR #75, opened 2023-04-17
Don't reselect NIDs after initialising a room

DMRobertson:
We fetch those NIDs above in `eventIDToNID` by adding a `RETURNING` clause to the `INSERT` statement.


DMRobertson:
Closing because I merged this into #71.

PR #84, opened 2023-04-25
If `bump_event_types` is specified, don't bump for RoomUpdates that aren't RoomEventUpdates.

### LINKED ISSUE #83: If `bump_event_types` is set, read receipts can still bump rooms to the top of the list
A bug in #34.

As reported by @stefanceriu, it seems that changing your own nickname or reacting in a room will bump that room to the top of its list, even if memberships and reactions are excluded by bump_event_types. Other users' memberships and reactions seem to be excluded correctly.

DMRobertson:
Fixes #83.

### DIFF
--- a/tests-e2e/lists_test.go
+++ b/tests-e2e/lists_test.go
@@ -2,6 +2,7 @@ package syncv3_test
 
 import (
 	"fmt"
+	"github.com/matrix-org/sliding-sync/sync3/extensions"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


kegsay (approved):
LGTM!

tests-e2e/lists_test.go
```diff
@@ -973,9 +974,35 @@ func TestBumpEventTypesHandling(t *testing.T) {
 	)
 
 	m.MatchResponse(t, bobRes, matchBobSeesRoom2Bumped)
-}
 
-// Test joining a room puts it at the top of your list. What about invites?
+	// The read receipt stuff here specifically checks for the bug in
+	// https://github.com/matrix-org/sliding-sync/issues/83
```
DMRobertson, line 979: I did try making this its own test but couldn't reproduce the behaviour. Having slept on it, I think the crucial thing is that there has to be some other update to the room which updates the LastModifiedTimestamp in the global caches. (Here, Charlie's join.)

PR #87, opened 2023-04-26
Support refreshing tokens

### LINKED ISSUE #51: Scope the connections to devices and not to a hash of the access token
It looks like some stuff is scoped to a hash of the access token, which will become a problem when we have refreshing access tokens, either via MSC2918 or with OIDC.

https://github.com/matrix-org/sliding-sync/blob/2c91fcb34b84821a71ac8d89da1756b047f2a59e/internal/request.go#L11-L24

It should be using the device_id instead, which are stable even when the token refreshes

Maybe related to #22?

Some (internal) discussions about that:

 - https://matrix.to/#/!TSgXVVnbcyPGrFniBq:connecteu.rs/$Bu_KKspZZD1oRW7i35jKH66mnSJz0WRODi29UZFHqZg?via=element.io&via=matrix.org&via=one.ems.host
 - https://matrix.to/#/!TSgXVVnbcyPGrFniBq:connecteu.rs/$PDCYL6WkAa7C5uN0j7a5J8e2X3Qw63DYC8IaBkBZTPs?via=element.io&via=matrix.org&via=one.ems.host

DMRobertson:
Closes #51.

VERY WIP

I am going to force push to this, review at your own peril


internal/request_test.go
```diff
@@ -1,24 +0,0 @@
-package internal
-
-import (
-	"net/http"
-	"testing"
-)
-
-func TestDeviceIDFromRequest(t *testing.T) {
```
DMRobertson, line 8: We can no longer directly determine the device id (i.e. the unique id used by the proxy to identify a device) by the information provided in the request.

internal/v2_devices.go
```diff
@@ -0,0 +1,19 @@
+package internal
+
+import "fmt"
+
+// TODO: should this live in package sync2?
```
DMRobertson, line 5: I think so: the sync3 tables are notionally dependent on sync2. (There aren't any foreign key constraints that express this, however.)

sync2/storage.go
```diff
@@ -177,8 +186,3 @@ func (s *Storage) UpdateDeviceSince(deviceID, since string) error {
 	_, err := s.db.Exec(`UPDATE syncv3_sync2_devices SET since = $1 WHERE device_id = $2`, since, deviceID)
 	return err
 }
-
-func (s *Storage) UpdateUserIDForDevice(deviceID, userID string) error {
-	_, err := s.db.Exec(`UPDATE syncv3_sync2_devices SET user_id = $1 WHERE device_id = $2`, userID, deviceID)
-	return err
-}
```
DMRobertson, line 184: We now need to know the user_id in order to form a device_id; prior to this it was possible that the user id had not yet been fetched from /whoami.

Come to think of it, we could make the user_id column non null here.

sync3/handler/handler.go
```diff
@@ -379,6 +358,38 @@ func (h *SyncLiveHandler) setupConnection(req *http.Request, syncReq *sync3.Requ
...
+		log.Warn().Err(err).Str("device_id", hsDeviceID).Msg("failed to get user ID from device ID")
+		return nil, &internal.HandlerError{
+			StatusCode: http.StatusBadGateway,
+			Err:        err,
+		}
+	}
+
+	// Either: create a brand-new row for this device, or
+	// update an existing row for this device with the latest access token.
+	// TODO: If the latter, we need to tell any existing poller about the new token.
```
DMRobertson, line 380: Highlighting this TODO

sync2/handler2/handler.go
```diff
@@ -139,6 +159,96 @@ func (h *Handler) StartV2Pollers() {
...
+// Otherwise this function
+//   - contacts the homeserver to fetch the "real" device ID,
+//   - builds the new device ID and updates the DB accordingly, and
+//   - updates the given sync2.Device with the new device ID.
+//
+// The DB updates for this device are updated atomically in a transaction.
+func (h *Handler) ensureDeviceIDMigrated(d sync2.Device) error {
+	if internal.IsNewDeviceID(d.DeviceID) {
+		return nil
+	}
```
DMRobertson, line 173: Need to write some kind of test for this

sync2/handler2/handler.go
```diff
@@ -383,13 +493,13 @@ func (h *Handler) OnLeftRoom(userID, roomID string) {
...
-	logger.Info().Str("user", p.UserID).Msg("EnsurePolling: new request")
+	logger.Info().Str("device", p.DeviceID).Msg("EnsurePolling: new request")
 	defer func() {
-		logger.Info().Str("user", p.UserID).Msg("EnsurePolling: request finished")
+		logger.Info().Str("device", p.DeviceID).Msg("EnsurePolling: request finished")
 	}()
 	dev, err := h.v2Store.Device(p.DeviceID)
 	if err != nil {
-		logger.Err(err).Str("user", p.UserID).Str("device", p.DeviceID).Msg("V3Sub: EnsurePolling unknown device")
+		logger.Err(err).Str("device", p.DeviceID).Msg("V3Sub: EnsurePolling unknown device")
```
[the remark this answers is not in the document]
DMRobertson: (But I didn't want to make this diff larger by renaming or removing fields. I still kinda want to make all the tables use a `user_id, actual_device_from_the_homeserver` pair, but I'll try to suppress that urge.)

sync2/handler2/handler.go
```diff
@@ -139,6 +159,96 @@ func (h *Handler) StartV2Pollers() {
...
+			"/whoami response was for the wrong user. Queried for %s, but got response for %s",
+			d.UserID, gotUserID,
+		)
+	}
+
+	newDeviceID := internal.ProxyDeviceID(gotUserID, gotHSDeviceID)
+	// Note: I'm assuming that h.Store.DB and h.v2Store.db point to the same DB
+	// so that I can do migrate across both stores in one transaction.
+	// TODO: I've put the DB writing logic here. It's icky that sync2 is mucking around
+	// with sync3's data. Maybe move this out to a new file, e.g. internal/migrations.go?
```
kegsay, line 192: Yes please.

sync2/storage.go
kegsay: Yes please.

sync3/handler/handler.go
```diff
@@ -280,20 +280,29 @@ func (h *SyncLiveHandler) setupConnection(req *http.Request, syncReq *sync3.Requ
...
+		log.Warn().Err(err).Msg("failed to get access token from request")
 		return nil, &internal.HandlerError{
-			StatusCode: 400,
+			StatusCode: http.StatusUnauthorized,
 			Err:        err,
 		}
 	}
 
+	v2device, err := h.V2Store.DeviceByPlaintextAccessToken(accessToken)
+	if err != nil {
```
kegsay, line 293: We don't log `err` at all, which feels like a mistake? It would be nice to not log if it's just because its the first time we've seen the token (ErrNoRows)

kegsay (commented):
This looks exactly as I'd expect.

sync2/storage.go
DMRobertson: Ahh, it already is null. It's just that sometimes we put the empty string into this column.

There are six entries in this table with `user_id = ''`. But they all have `since = ''` too. I think we ought to drop such devices from the table.

sync3/handler/handler.go
DMRobertson: See 420d58a

internal/v2_devices.go
DMRobertson: Moved to the new migrations.go file.

DMRobertson:
Closing in favour of newer PRs, #90 for starters.

Issue #89, opened 2023-04-27
Decouple access token tracking from device tracking


DMRobertson:
A step towards #51.

```sql
CREATE TABLE syncv3_sync2_tokens (
    token_hash TEXT NOT NULL PRIMARY KEY, -- SHA256(access token)
    user_id TEXT NOT NULL,
    device_id TEXT NOT NULL,  -- FK constraint to devices table?
    last_seen TIMESTAMP NOT NULL,
    token_encrypted TEXT NOT NULL
)
```
Use:
- When a new conn is made, lookup the (user, device) pair by token_hash.
  - Use this pair as the key within the proxy, to avoid confusing ourselves
  - If unknown, call /whoami to determine the (user, device) pair.
- Update the last_seen timestamp to now.

Rationale:

- Without separate token tracking, we, would have to assume one device has at most one access token in play. This opens us up to race conditions.
- Longer term, we want to support the prospect of a "long-lived token", or even making the proxy request tokens on the client's behalf.

Implementation:

- One-time migrate the existing access tokens from devices table. This requires many /whoami calls to fetch the actual device ID from the homeserver.
- Periodically clean out access tokens (from cache and DB). that haven't been used lately, say in a week. We would keep rows in the device table around though, for the `since` parameter in particular.
- Optimisation: cache the tokens table in memory. We'll have to make sure that any invalidations apply to the cache as well as the DB.
- Optimisation: only write the last_seen timestamp with coarse resolution, e.g. 24hr.

Questions:
- If the poller learns that a token has expired, should we record that fact? Otherwise we might expire a token only for the client to use it, meaning we'll make /whoami calls. Maybe a `expired BOOLEAN NOT NULL` field?

kegsay:
> If the poller learns that a token has expired, should we record that fact? Otherwise we might expire a token only for the client to use it, meaning we'll make /whoami calls. Maybe a expired BOOLEAN NOT NULL field?

Probably not as it hurts self-healing. If Synapse returns a 401/403 for some _other_ reason and we mark the token as expired, _but it isn't really_, then we break the client until they refresh their token. If instead we don't mark it as expired, then yes we will stop polling due to the 401/403, but they can "kick" the server by using the token again, which will then hopefully return 200 OK when used again.

kegsay:
> Optimisation: cache the tokens table in memory. We'll have to make sure that any invalidations apply to the cache as well as the DB.

[...]

Issue #36, opened 2023-03-21
Cypress test `should send unsubscribe_rooms for every room switch` seems flakey


DMRobertson:
e.g. https://github.com/matrix-org/sliding-sync/actions/runs/4482864263/jobs/7881442972

```
      AssertionError: room_subscriptions is object: expected undefined to be an object
```

kegsay:
Related: #42 ? Need to check video.

DMRobertson:
And https://github.com/matrix-org/sliding-sync/actions/runs/4553358923/jobs/8029820210 , which should now have a video at https://github.com/matrix-org/sliding-sync/actions/runs/4553358923

DMRobertson:
And https://github.com/matrix-org/sliding-sync/actions/runs/4613554076/jobs/8155657605#step:9:2435

DMRobertson:
I spent an hour or so looking into this. From what I can see, the react/js sdk maintains an explicit subscription to whichever room is open in the client. (You still want to see updates that room, even if it's outside of your sliding window(s).)

AFAICS The `room_subscriptions` field is populated here, whenever `newSubscriptions` is nonempty.

https://github.com/matrix-org/matrix-js-sdk/blob/c805b7e29d6c8fd592112df8acc7ea78c532db5e/src/sliding-sync.ts#L850-L866

`newSubscriptions` is the set difference `this.desiredRoomSubscriptions - this.confirmedRoomSubscriptions`. So the room we're looking for is either not desired, already confirmed, or both when we see this failure mode.

I guess the next step would be to get the SDK to log those sets and somehow get CI to expose those logs to us.

DMRobertson:
I haven't seen this fail in a while. Maybe something got fixed in the react/js sdk?

DMRobertson:
Let's assume this magically unflaked itself, and reopen if not.

DMRobertson:
> Let's assume this magically unflaked itself, and reopen if not.

Spoke too soon. https://github.com/matrix-org/sliding-sync/actions/runs/5598990176/jobs/10239377271

PR #90, opened 2023-04-27
OIDC: track tokens separately to devices

DMRobertson:
This is the first part of my plan proposed in https://github.com/matrix-org/sliding-sync/issues/89#issuecomment-1527909808. Note that it targets a separate `dmr/oidc` branch.

This diff

- introduces a v2 tokens table,
- changes the schema of the v2 devices table,
- changes the sync2 part of the application to uses the tables, and
- passes (user, device) pairs around as conn and poller IDs, including to the sync3 part of the application

It does not
- alter the v3 parts of the schema (want to use actual device IDs and to include user IDs), nor
- add any sort of migration.


pubsub/v2.go
```diff
@@ -78,6 +78,7 @@ type V2InitialSyncComplete struct {
 func (*V2InitialSyncComplete) Type() string { return "V2InitialSyncComplete" }
 
 type V2DeviceData struct {
+	UserID   string
 	DeviceID string
 	Pos      int64
```
DMRobertson, line 83: Is this actually Pos or is it a since token?

DMRobertson: It's a position in the `syncv3_device_data_seq` sequence.

sync2/handler2/handler.go
```diff
@@ -150,12 +154,11 @@ func (h *Handler) OnTerminated(userID, deviceID string) {
...
-	h.Store.ToDeviceTable.DeleteAllMessagesForDevice(deviceID)
-	h.Store.DeviceDataTable.DeleteDevice(userID, deviceID)
-	// also notify v3 side so it can remove the connection from ConnMap
+func (h *Handler) OnExpiredToken(accessTokenHash, userID, deviceID string) {
+	h.v2Store.TokensTable.Delete(accessTokenHash)
+	// Notify v3 side so it can remove the connection from ConnMap
 	h.v2Pub.Notify(pubsub.ChanV2, &pubsub.V2ExpiredToken{
+		UserID:   userID,
 		DeviceID: deviceID,
 	})
```
DMRobertson, line 163: At first I didn't delete the token, but then the token would get selected on incoming requests only for the homeserver to reject it. Loop of doom.

Do we care about deleting messages etc? We could have some kind of cleanup that nukes messages for devices if they've had no access tokens for at least e.g. 30 days.

sync2/poller.go
```diff
@@ -151,15 +156,21 @@ func (h *PollerMap) NumPollers() (count int) {
...
 		go h.execute()
 	}
-	poller, ok := h.Pollers[deviceID]
+	poller, ok := h.Pollers[pid]
 	// a poller exists and hasn't been terminated so we don't need to do anything
 	if ok && !poller.terminated.Load() {
+		if poller.accessToken != accessToken {
+			logger.Warn().
+				Str("user_id", pid.UserID).
+				Str("device_id", pid.DeviceID).
```
DMRobertson, line 171: Redundant, this is set on the logger by the caller

sync2/poller.go
```diff
@@ -169,28 +180,31 @@ func (h *PollerMap) EnsurePolling(accessToken, userID, deviceID, v2since string,
...
 	go poller.Poll(v2since)
-	h.Pollers[deviceID] = poller
+	h.Pollers[pid] = poller
 
 	h.pollerMu.Unlock()
 	if needToWait {
 		poller.WaitUntilInitialSync()
 	} else {
-		logger.Info().Str("user", userID).Msg("a poller exists for this user; not waiting for this device to do an initial sync")
+		logger.Info().Str("user", poller.userID).Msg("a poller exists for this user; not waiting for this device to do an initial sync")
```
DMRobertson, line 207: `user` field set by caller

sync2/tokens_table.go
```diff
@@ -0,0 +1,233 @@
...
+			"token (hash %s) found with user+device mismatch: got (%s, %s), expected (%s, %s)",
+			tokenHash, gotUserID, gotDeviceID, userID, deviceID,
+		)
+	}
+	accessToken, err = t.decrypt(encToken)
+	return
+}
+
+// Delete looks up a token by its hash and deletes the row.
+// TODO: what happens if the hash isn't found?
```
DMRobertson, line 226: Probably return NoRowsError and log as a warning.

sync2/tokens_table_test.go
```diff
@@ -0,0 +1,16 @@
+package sync2
+
+import (
+	"testing"
+)
+
+// Sanity check that different tokens have different hashes
```
DMRobertson, line 7: Porting this from `request_test.go`

sync3/connmap.go
```diff
@@ -71,7 +71,7 @@ func (m *ConnMap) CreateConn(cid ConnID, newConnHandler func() ConnHandler) (*Co
 	conn = NewConn(cid, h)
 	m.cache.Set(cid.String(), conn)
 	m.connIDToConn[cid.String()] = conn
```
DMRobertson, line 73: maybe this should now be a map from ConnID structs, not strings?

sync3/handler/handler.go
```diff
@@ -276,25 +281,53 @@ func (h *SyncLiveHandler) serve(w http.ResponseWriter, req *http.Request) error
...
+			}
+		}
+	}
+	log := hlog.FromRequest(req).With().Str("user", token.UserID).Str("device", token.DeviceID).Logger()
+
+	// Record the fact that we've recieved a request from this token
+	err = h.V2Store.TokensTable.MaybeUpdateLastSeen(token, time.Now())
+	if err != nil {
+		// Not fatal---log and continue.
+		log.Warn().Msg("Unable to update last seen timestamp")
```
DMRobertson, line 320: Include the `err` here

tests-integration/connection_test.go
```diff
@@ -639,17 +639,10 @@ func TestExpiredAccessToken(t *testing.T) {
 	})
 	// now expire the token
 	v2.invalidateToken(aliceToken)
-	// now do another request, this should 400 as it expires the session
+	// now do another request, this should 401
 	req := sync3.Request{}
 	req.SetTimeoutMSecs(1)
 	_, body, statusCode := v3.doV3Request(t, context.Background(), aliceToken, res.Pos, req)
```
DMRobertson, line 645: Wasn't sure if this change in behaviour is desirable?

tests-integration/v3_test.go
```diff
@@ -97,6 +110,12 @@ func (s *testV2Server) userID(token string) string {
 	return s.tokenToUser[token]
 }
 
+func (s *testV2Server) deviceID(token string) string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.tokenToDevice[token]
+}
```
DMRobertson, line 117: I wonder if this is safe. Should the mutex guard the (user, device) pair?

sync2/tokens_table.go
kegsay: It's not great to return errors when nothing was deleted, as it isn't a fatal condition and is more likely to cause issues as people forget to check `err` for this specific type of error.

sync2/tokens_table.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (t *TokensTable) GetTokenAndSince(userID, deviceID, tokenHash string) (accessToken, since string, err error) {
+	var encToken, gotUserID, gotDeviceID string
+	row := t.db.QueryRow(`
+		SELECT token_encrypted, since, user_id, device_id
+		FROM syncv3_sync2_tokens JOIN syncv3_sync2_devices USING (user_id, device_id)
+		WHERE token_hash = $1;
+	`, tokenHash)
+	err = row.Scan(&encToken, &since, &gotUserID, &gotDeviceID)
```
kegsay, line 211: You'd typically inline this e.g `if err := QueryRow(...).Scan(&foo); err != nil { ... }`.

sync2/tokens_table.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return
+	}
+	if gotUserID != userID || gotDeviceID != deviceID {
```
kegsay, line 215: We definitely need the inserts in `syncv3_sync2_tokens` and `syncv3_sync2_devices` to be in a txn then if we are going to return an error if they aren't set.

sync2/tokens_table.go
```diff
@@ -0,0 +1,233 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	token.AccessToken = plaintextToken
+	token.AccessTokenHash = tokenHash
+	return &token, nil
+}
+
+type TokenWithSince struct {
```
kegsay, line 128: Bad name as it is awkward when we want to add additional fields to this struct beyond "since".

kegsay (approved):
Good so far; please add unit tests for new tables.

DMRobertson:
> please add unit tests for new tables.

There were some, but I've pulled some of this out to a separate file in 5404893.

sync3/handler/handler.go
```diff
@@ -366,19 +367,58 @@ func (h *SyncLiveHandler) setupConnection(req *http.Request, syncReq *sync3.Requ
...
+			}
+		}
+		log.Warn().Err(err).Msg("failed to get user ID from device ID")
+		return nil, &internal.HandlerError{
+			StatusCode: http.StatusBadGateway,
+			Err:        err,
+		}
+	}
+
+	// TODO: should the two inserts be wrapped in a transaction?
```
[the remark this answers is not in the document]
DMRobertson: Done in 5c9afec.

sync2/tokens_table.go
DMRobertson: Thanks. I've made 14cf838 log a warning but otherwise ignored this scenario.

sync2/tokens_table.go
DMRobertson: I wasn't happy with it (but also I find it relatively easy to rename things---IntelliJ's/Goland's refactoring tools seem to work very well for go.

DMRobertson:
This seems happy other than #36. Going to merge into the oidc branch.

PR #98, opened 2023-05-03
Migrate existing data to new schema on startup

DMRobertson:
Tested sucessfully against an isolated backup of the m.org proxy deployment. (Circa 5 minutes to run and commit.)

I would like to write a few test cases (see https://github.com/matrix-org/sliding-sync/issues/89#issuecomment-1527909808) before closing #89 and #51.

### DIFF
--- a/cmd/testmigration/main.go
+++ b/cmd/testmigration/main.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"github.com/jmoiron/sqlx"
+	"github.com/matrix-org/sliding-sync/sync2"
+	"net/http"
+	"os"
+	"time"
+)
+
+const (
+	// Required fields
+	EnvServer = "SYNCV3_SERVER"
+	EnvDB     = "SYNCV3_DB"
+	EnvSecret = "SYNCV3_SECRET"
+
+	// Migration test only
+	EnvMigrationCommit = "SYNCV3_TEST_MIGRATION_COMMIT"
+)
+
+func main() {
+	args := map[string]string{
+		EnvServer:          os.Getenv(EnvServer),
+		EnvDB:              os.Getenv(EnvDB),
+		EnvSecret:          os.Getenv(EnvSecret),
+		EnvMigrationCommit: os.Getenv(EnvMigrationCommit),
+	}
+
+	db, err := sqlx.Open("postgres", args[EnvDB])
+	if err != nil {
+		panic(err)
+	}
+
+	v2Client := &sync2.HTTPClient{
+		Client: &http.Client{
+			Timeout: 5 * time.Minute,
+		},
+		DestinationServer: args[EnvServer],
+	}
+
+	err = sync2.MigrateDeviceIDs(db, args[EnvDB], v2Client, args[EnvMigrationCommit] != "")
+	if err != nil {
+		panic(err)
+	}
+}
--- a/cmd/testmigration/main.go
+++ b/cmd/testmigration/main.go
@@ -38,7 +38,7 @@ func main() {
 		DestinationServer: args[EnvServer],
 	}
 
-	err = sync2.MigrateDeviceIDs(db, args[EnvDB], v2Client, args[EnvMigrationCommit] != "")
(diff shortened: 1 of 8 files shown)


sync2/migration.go
```diff
@@ -0,0 +1,264 @@
+package sync2
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"github.com/jmoiron/sqlx"
+	"github.com/matrix-org/sliding-sync/sqlutil"
+)
+
+func MigrateDeviceIDs(db *sqlx.DB, secret string, whoamiClient Client, commit bool) error {
```
DMRobertson, line 10: This isn't called at present. I'll need to write some logic that tries to actually do the migration on startup.

sync2/migration.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	logger.Info().Msgf("Got %d devices to migrate", len(devices))
+
+	hasher := sha256.New()
+	hasher.Write([]byte(secret))
+	key := hasher.Sum(nil)
+
+	// TODO: can we get away with doing this sequentially, or should we parallelise
+	//       this like the poller startup routine does?
```
kegsay, line 130: This is slower but probably fine?

sync2/migration.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func expectOneRowAffected(ra int64) bool          { return ra == 1 }
+func expectAnyNumberOfRowsAffected(ra int64) bool { return true }
+func expectAtMostOneRowAffected(ra int64) bool    { return ra == 0 || ra == 1 }
+
+func finish(txn *sqlx.Tx) (err error) {
+	_, err = txn.Exec(`
+		ALTER TABLE syncv3_sync2_devices
+		DROP COLUMN v2_token_encrypted,
```
kegsay, line 240: Why drop this?

DMRobertson: I thought the point was to move the tokens to the tokens table. Is there any reason to keep it around in the devices table? I suppose that could reflect that access token that a particular poller is using right now in the moment... but I'm not sure how we'd use that information.

sync2/migration.go
DMRobertson: Yeah. Just trying to keep things simple for now.

DMRobertson: On a test copy of the matrix.org deployment's database, this took <5min. I think that's more than fine.

state/to_device_table.go
```diff
@@ -81,7 +81,12 @@ func (t *ToDeviceTable) DeleteMessagesUpToAndIncluding(userID, deviceID string,
 }
 
 func (t *ToDeviceTable) DeleteAllMessagesForDevice(userID, deviceID string) error {
+	// TODO: should these deletes take place in a transaction?
```
DMRobertson, line 84: I don't think this is crucial?

sync2/migration.go
```diff
@@ -0,0 +1,370 @@
...
+		ALTER TABLE syncv3_to_device_messages
+		ALTER COLUMN user_id SET NOT NULL;
+	`)
+	if err != nil {
+		return
+	}
+
+	_, err = txn.Exec(`
+		ALTER TABLE syncv3_to_device_ack_pos
+		ALTER COLUMN user_id SET NOT NULL,
```
DMRobertson, line 353: The cleanup for syncv3_to_device_ack_pos above ensures that this ALTER COLUMN succeeds.

cmd/testmigration/main.go
DMRobertson, line 1: It may be useful to keep this around---this gives us a way to run the migration manually without having to startup the proxy proper.

sync2/migration.go
kegsay: I just didn't see a tokens table when I looked.

DMRobertson: Oh bugger, I forgot to insert the new rows! Arg.

kegsay (approved):

DMRobertson:
I'm going to merge this as it is. I would like to spend a little more time seeing if there are any easy speedups, to minimise downtime. ATM the slowest parts are updating the to_device table. There are occasionally slow /whoami responses too.

PR #100, opened 2023-05-09
readme: Add note for advertising support via `.well-known`

bnjbvr:
This was the step I missed during the set up of the sliding-sync proxy for my server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ $ SYNCV3_SECRET=$(cat .secret) SYNCV3_SERVER="https://matrix-client.matrix.org" 
 ```
 Using a Docker image:
 ```
-docker run --rm -e "SYNCV3_SERVER=https://matrix-client.matrix.org" -e "SYNCV3_SECRET=$(cat .secret)" -e "SYNCV3_BINDADDR=:8008" -e "SYNCV3_DB=user=$(whoami) dbname=syncv3 sslmode=disable host=host.docker.internal" -p 8008:8008 ghcr.io/matrix-org/sliding-sync:v0.98.0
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ $ SYNCV3_SECRET=$(cat .secret) SYNCV3_SERVER="https://matrix-client.matrix.org"
 ```
 Using a Docker image:
 ```
-docker run --rm -e "SYNCV3_SERVER=https://matrix-client.matrix.org" -e "SYNCV3_SECRET=$(cat .secret)" -e "SYNCV3_BINDADDR=:8008" -e "SYNCV3_DB=user=$(whoami) dbname=syncv3 sslmode=disable host=host.docker.internal" -p 8008:8008 ghcr.io/matrix-org/sliding-sync:v0.98.0
+docker run --rm -e "SYNCV3_SERVER=https://matrix-client.matrix.org" -e "SYNCV3_SECRET=$(cat .secret)" -e "SYNCV3_BINDADDR=:8008" -e "SYNCV3_DB=user=$(whoami) dbname=syncv3 sslmode=disable host=database.host:5432" -p 8008:8008 ghcr.io/matrix-org/sliding-sync:v0.98.0
```
kegsay, line 44: `host.docker.internal` is correct if you have a local postgres, as per the earlier `createdb` command. Why did you change this, when `database.host` is always wrong? I don't think it's any clearer, given this is the `host` key of the `SYNCV3_DB` env var, so not exactly unclear that it is the db host?

kegsay (changes requested):
Otherwise LGTM, thanks!

kegsay (approved):

PR #106, opened 2023-05-10
Ensure poller restarts with new access token after the old one expires

DMRobertson:
This change includes an integration test which checks that 

- sliding sync requests after using a refreshed access token give sensible data, and that
- the poller uses a previous `since` token when switching to the new access token.

It also includes changes to make this test pass, plus drive-by comments.


sync3/handler/handler.go
```diff
@@ -704,6 +704,7 @@ func (h *SyncLiveHandler) OnAccountData(p *pubsub.V2AccountData) {
 }
 
 func (h *SyncLiveHandler) OnExpiredToken(p *pubsub.V2ExpiredToken) {
+	h.V3Pub.OnTokenExpired(p)
```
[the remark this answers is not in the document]
kegsay: Happy to rename the field to `EnsurePoller`.

sync3/handler/ensure_polling.go
```diff
@@ -93,6 +109,38 @@ func (p *EnsurePoller) OnInitialSyncComplete(payload *pubsub.V2InitialSyncComple
...
+	}
+	// There's a potential race here. Suppose that
+	//
+	//  - a v3 request arrives for a brand-new device
+	//  - homeserver recognises the access token in the /whoami call
+	//  - we call EnsurePolling for that token
+	//  - the poller makes an initial sync and learns that the token has expired
+	//
+	// If we close the channel, the EnsurePolling call that's blocking on it will
+	// continue as-if there was initial sync data to receive. But there isn't.
```
kegsay, line 129: How is this a race exactly? We're just unblocking the v3 request on expired tokens.

tests-integration/token_management_test.go
```diff
@@ -0,0 +1,120 @@
...
+				events: []json.RawMessage{bobMsg},
+			}),
+		},
+		NextBatch: "after_alice_incremental_poll",
+	})
+
+	t.Log("Alice makes a new sliding sync connection with her new token")
+	resChan := make(chan *sync3.Response)
+	go func() {
+		resChan <- v3.mustDoV3Request(t, aliceToken2, req)
```
kegsay, line 102: Why do this concurrently?

kegsay (approved):
Excellent.

tests-integration/token_management_test.go
DMRobertson: I was trying to ensure that this v3 sync blocked on the v2 incremental poll. But I think checking for `bobMsg` below does that anyway. Will tidy up.

sync3/handler/ensure_polling.go
```diff
@@ -33,6 +44,11 @@ func NewEnsurePoller(notifier pubsub.Notifier) *EnsurePoller {
 func (p *EnsurePoller) EnsurePolling(pid sync2.PollerID, tokenHash string) {
 	p.mu.Lock()
 	// do we need to wait?
+	// TODO: this lookup is based on (user, device) pair. If the same user logs in on
+	// another device, we will wait for the poller to make an initial sync. We could do
```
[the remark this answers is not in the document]
DMRobertson: I'd forgotten there was logic for this in the V2 side. Will fixup.

sync3/handler/ensure_polling.go
DMRobertson: Maybe not a race; this just came to mind as a potentially bad outcome. My concern was that the proxy might assume that when the v3 request is unblocked it has some kind of sync v2 data persisted to DB. But I suppose it has to anticipate and handle poller failures anyway?

sync3/handler/ensure_polling.go
```diff
@@ -93,6 +104,21 @@ func (p *EnsurePoller) OnInitialSyncComplete(payload *pubsub.V2InitialSyncComple
...
+	pid := sync2.PollerID{UserID: payload.UserID, DeviceID: payload.DeviceID}
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	pending, exists := p.pendingPolls[pid]
+	if !exists {
+		// We weren't tracking the state of this poller, so we have nothing to clean up.
+		return
+	}
+	if pending.ch != nil {
+		close(pending.ch)
```
kegsay, line 117:
```suggestion
		// unblock anything waiting on this sync, it's never going to happen now.
		close(pending.ch)
```

tests-integration/token_management_test.go
```diff
@@ -0,0 +1,191 @@
...
+	if code != 400 {
+		t.Errorf("got HTTP %d want 400", code)
+	}
+	if gjson.ParseBytes(body).Get("errcode").Str != "M_UNKNOWN_POS" {
+		t.Errorf("got %v want errcode=M_UNKNOWN_POS", string(body))
+	}
+
+	t.Log("Prepare to tell a poller using aliceToken2 that Alice created a room and that Bob joined it.")
+	bobMsg := testutils.NewMessageEvent(t, bob, "Hello, world!")
+	v2.queueResponse(aliceToken2, sync2.SyncResponse{
```
kegsay, line 92: It would be good to also check that if you queueResponse BEFORE  you do a sync with the new token that that data is remembered (and we don't say, drop it or request a 0 limit on it like we do with normal initial syncs).

kegsay (approved):

kegsay:
Let's land this before we add anything more to this PR.

tests-integration/token_management_test.go
DMRobertson: To check, do you mean that call to queueResponse to use the old token?

sync3/handler/ensure_polling.go
DMRobertson: Missed this, will include in a future PR

PR #107, opened 2023-05-11
bugfix: don't corrupt the user cache

kegsay:
Temporary fix for a bug which could cause the proxy to send back incorrect timeline events, and miss some entirely.

Repro:
- initial load with msg A
- live stream B C
- add a room sub with limit=5 -> we return [A] due to using the old loadPos and nuke B,C in the process
- live stream D E -> gets added to [A] forming [A,D,E] which is wrong.

This is only a temp fix because in reality we want to be using `roomToLoadPos map[string]int64` everywhere from now on. The current fix causes `UserRoomData` responses which have the correct timeline but incorrect (newer) other fields like `IsInvite`, `NotificationCount`, etc.

### DIFF
--- a/sync3/caches/user.go
+++ b/sync3/caches/user.go
@@ -363,6 +363,7 @@ func (c *UserCache) LazyLoadTimelines(ctx context.Context, loadPos int64, roomID
 		if !ok {
 			urd = NewUserRoomData()
 		}
+		oldLoadPos := urd.LoadPos
 		urd.Timeline = events
 		urd.LoadPos = loadPos
 		if len(events) > 0 {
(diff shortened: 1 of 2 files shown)


sync3/caches/user.go
```diff
@@ -371,7 +372,13 @@ func (c *UserCache) LazyLoadTimelines(ctx context.Context, loadPos int64, roomID
 		}
 
 		result[roomID] = urd
-		c.roomToData[roomID] = urd
+		// only replace our knowledge if it is the future
+		// TODO FIXME: this is an incomplete solution as all the non-Timeline fields we return in this
+		// function will be AHEAD of the load pos provided, meaning you could see a stale timeline with
+		// incorrect notif counts (until you've consumed the channel)
```
DMRobertson, line 378: As discussed: this is unfortunate, but:
- it will self-heal shortly as the channel of updates is consumed
- it is a less-bad failure mode than missing messages!

DMRobertson (approved):

PR #108, opened 2023-05-11
bugfix: ensure global metadata is consistently made correctly

kegsay:
On startup it was possible for it to miss ChildSpaceRooms which would then cause a nil panic. With regression test.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/getsentry/sentry-go"
 	"os"
 	"strings"
+
+	"github.com/getsentry/sentry-go"
 
 	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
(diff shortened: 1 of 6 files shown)


state/storage.go
```diff
@@ -766,9 +767,9 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx) (result map[string][]string, me
 	}
 	metadata = make(map[string]internal.RoomMetadata)
 	for roomID, joinedMembers := range result {
-		metadata[roomID] = internal.RoomMetadata{
-			JoinCount: len(joinedMembers),
-		}
+		m := internal.NewRoomMetadata(roomID)
+		m.JoinCount = len(joinedMembers)
+		metadata[roomID] = *m
```
kegsay, line 772: This is the actual fix, as this didn't make `ChildSpaceRooms`

sync3/caches/user.go
```diff
@@ -164,17 +164,16 @@ func (i *InviteData) RoomMetadata() *internal.RoomMetadata {
 	if i.RoomType != "" {
 		roomType = &i.RoomType
 	}
-	return &internal.RoomMetadata{
```
DMRobertson, line 167: Was it a problem that we didn't make a ChildSpaceRooms map here, too? Ditto for the two hunks below.

DMRobertson (approved):

sync3/caches/user.go
kegsay: Nope as they were stubs and not used by the global cache

PR #113, opened 2023-05-16
Capture, report and reraise panics in `Listen`

### LINKED ISSUE #109: Sentry doesn't report panics which crash the process
i.e https://github.com/matrix-org/sliding-sync/pull/108

We should, because crashes are bad and sentry worthy ;)

DMRobertson:
Closes #109.

Tested by

- adding a panic into OnNewEvent, like in #109, checking that we don't get a Sentry event
- adding the new code, checking that we do get a Sentry event

### DIFF
--- a/internal/sentry.go
+++ b/internal/sentry.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"github.com/getsentry/sentry-go"
+	"time"
 )
 
 // GetSentryHubFromContextOrDefault is a version of sentry.GetHubFromContext which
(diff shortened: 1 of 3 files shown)


sync3/handler/handler.go
```diff
@@ -41,13 +41,13 @@ var logger = zerolog.New(os.Stdout).With().Timestamp().Logger().Output(zerolog.C
...
-	Storage    *state.Storage
-	V2Store    *sync2.Storage
+	V2           sync2.Client
+	Storage      *state.Storage
+	V2Store      *sync2.Storage
 	V2Sub        *pubsub.V2Sub
 	EnsurePoller *EnsurePoller
 	ConnMap      *sync3.ConnMap
-	Extensions *extensions.Handler
+	Extensions   *extensions.Handler
```
DMRobertson, line 50: I guess this is just GoLand formatting a file that I touched?

internal/sentry.go
```diff
@@ -22,3 +23,20 @@ func GetSentryHubFromContextOrDefault(ctx context.Context) *sentry.Hub {
 	}
 	return hub
 }
+
+// ReportPanicsToSentry checks for panics by calling recover, reports any panic found to
+// sentry, and then reraises the panic. To have tracebacks included in the report, make
+// sure you call panic with an error, not a string.
```
kegsay, line 29: Errors are basically strings, so this doesn't make sense.

sync3/handler/handler.go
kegsay: Yes this is `go fmt`

kegsay (approved):

internal/sentry.go
DMRobertson: Poorly worded, but I have in mind this logic: https://github.com/getsentry/sentry-go/blob/eec094e9470dd3855eaf47b025d853bcbc13df68/client.go#L438-L447

Issue #85, opened 2023-04-25
Unexplained panics in `RemoveHero`


DMRobertson:
Internal sentry links: https://sentry.tools.element.io/organizations/element/issues/69588/?project=56 

```
runtime.errorString: runtime error: invalid memory address or nil pointer dereference
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/conn.go", line 97, in (*Conn).tryRequest.func1
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/internal/roomname.go", line 63, in (*RoomMetadata).RemoveHero
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/connstate.go", line 480, in (*ConnState).getInitialRoomData
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/connstate.go", line 429, in (*ConnState).buildRooms
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/connstate.go", line 157, in (*ConnState).onIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/connstate.go", line 126, in (*ConnState).OnIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/conn.go", line 109, in (*Conn).tryRequest
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/conn.go", line 204, in (*Conn).OnIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/handler.go", line 235, in (*SyncLiveHandler).serve
  File "github.com/matrix-org/sliding-sync@v0.99.2-0.20230414104506-c6f7516c80c9/sync3/handler/handler.go", line 166, in (*SyncLiveHandler).ServeHTTP
  File "net/http/server.go", line 2109, in HandlerFunc.ServeHTTP
[...]

PR #114, opened 2023-05-16
See if #85 is triggered by tests

DMRobertson:


DMRobertson:
On second thoughts, reopening. I'd like to run the proxy with this assertion and see if we hit it. If so that would confirm that a) my hypothesis in https://github.com/matrix-org/sliding-sync/issues/85#issue-1683543826 is correct, and b) my interpretation of the sentry tracebacks in https://github.com/matrix-org/sliding-sync/issues/85#issuecomment-1522135939 is correct.

kegsay:
`internal.Assert` only panics in debug mode, so maybe just test with a direct `panic("oh no")`

DMRobertson:
The cause is understood. New PR to follow with a test that reproduces the bug.

Issue #76, opened 2023-04-17
E2E flake `TestSecurityLiveStreamEventLeftLeak`


DMRobertson:
From https://github.com/matrix-org/sliding-sync/actions/runs/4721542126/jobs/8374942609?pr=74#step:7:689 (https://github.com/matrix-org/sliding-sync/pull/74)

```
❌ TestSecurityLiveStreamEventLeftLeak (360ms)
      client_test.go:788: POST localhost/_matrix/client/v3/register => 200 OK (19.685156ms)
      client_test.go:788: POST localhost/_matrix/client/v3/register => 200 OK (14.339205ms)
      client_test.go:788: POST localhost/_matrix/client/v3/createRoom => 200 OK (73.348872ms)
      client_test.go:788: POST localhost/_matrix/client/v3/join/!eXYomWUIWrztYepPVI:synapse => 200 OK (40.514845ms)
      client_test.go:788: POST localhost/_matrix/client/unstable/org.matrix.msc3575/sync => 200 OK (54.401836ms)
      client_test.go:788: POST localhost/_matrix/client/unstable/org.matrix.msc3575/sync => 200 OK (43.617632ms)
      client_test.go:788: POST localhost/_matrix/client/v3/rooms/!eXYomWUIWrztYepPVI:synapse/kick => 200 OK (18.835132ms)
      client_test.go:788: PUT localhost/_matrix/client/v3/rooms/!eXYomWUIWrztYepPVI:synapse/state/m.room.name/ => 200 OK (50.510826ms)
      security_test.go:70: SendEventSynced waiting for event ID $b024Oq4aIo78Y1uJn200cFMqw32Krb92vG9nOfTAsFk
      client_test.go:788: GET localhost/_matrix/client/v3/sync => 200 OK (28.141595ms)
      client_test.go:788: POST localhost/_matrix/client/unstable/org.matrix.msc3575/sync => 200 OK (1.868252ms)
      security_test.go:80: SlidingSyncUntil: tested response but it failed with: waiting for more messages, got 1
      client_test.go:788: POST localhost/_matrix/client/unstable/org.matrix.msc3575/sync => 200 OK (5.66166ms)
      security_test.go:125: MatchResponse: MatchRoomSubscription: MatchRoomTimelineMostRecent: timeline length mismatch: got 1 want at least 2
[...]

PR #115, opened 2023-05-17
Remove timeline cache; fix flakey test

kegsay:
- Fixes https://github.com/matrix-org/sliding-sync/issues/76 flakey test TestSecurityLiveStreamEventLeftLeak, caused by the test expecting 2 events to arrive in a single HTTP response.
- Remove the timeline cache in the user cache. This has been the source of numerous bugs around misordered and missing timeline events, and needs to be rewritten to take into account per-room load positions and handle trickling timeline limits better (starting with limit=0/1 then increasing to 20+).

### DIFF
--- a/sync3/caches/user.go
+++ b/sync3/caches/user.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/getsentry/sentry-go"
 
-	lru "github.com/hashicorp/golang-lru"
 	"github.com/matrix-org/sliding-sync/internal"
 	"github.com/matrix-org/sliding-sync/state"
 	"github.com/tidwall/gjson"
@@ -33,12 +32,12 @@ type UserRoomData struct {
 	HasLeft           bool
 	NotificationCount int
 	HighlightCount    int
-	// (event_id, last_event_id) -> closest prev_batch
-	// We mux in last_event_id so we can invalidate prev batch tokens for the same event ID when a new timeline event
-	// comes in, without having to do a SQL query.
-	PrevBatches       *lru.Cache
-	Timeline          []json.RawMessage
 	Invite            *InviteData
+
+	// these fields are set by LazyLoadTimelines and are per-function call, and are not persisted in-memory.
+	RequestedPrevBatch string
+	RequestedTimeline  []json.RawMessage
+
 	CanonicalisedName string // stripped leading symbols like #, all in lower case
 	// Set of spaces this room is a part of, from the perspective of this user. This is NOT global room data
 	// as the set of spaces may be different for different users.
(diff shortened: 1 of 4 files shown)


sync3/caches/user.go
```diff
@@ -287,97 +258,32 @@ func (c *UserCache) OnRegistered(ctx context.Context, _ int64) error {
...
-		urd, ok := c.roomToData[roomID]
+	for _, requestedRoomID := range roomIDs {
+		events := roomIDToEvents[requestedRoomID]
+		urd, ok := c.roomToData[requestedRoomID]
 		if !ok {
 			urd = NewUserRoomData()
 		}
-		oldLoadPos := urd.LoadPos
-		urd.Timeline = events
+		urd.RequestedTimeline = events
```
DMRobertson, line 280: Above we say that the new fields `RequestedTimeline` and `RequestedPrevBatch`

> are set by LazyLoadTimelines and are per-function call, and are not persisted in-memory.

But this `urd` comes from `c.roomToData` on a UserCache `c`, which sounds like it might be persistent. What's going on here?

DMRobertson: Ahh, `roomToData` holds UserRoomData values, so I am guessing that the local `urd` is a copy of (not a reference to) the cached struct.

sync3/caches/user.go
```diff
@@ -594,11 +500,7 @@ func (c *UserCache) OnSpaceUpdate(ctx context.Context, parentRoomID, childRoomID
 func (c *UserCache) OnNewEvent(ctx context.Context, eventData *EventData) {
 	// add this to our tracked timelines if we have one
 	urd := c.LoadRoomData(eventData.RoomID)
-	if len(urd.Timeline) > 0 {
-		// we're tracking timelines, add this message too
-		urd.Timeline = append(urd.Timeline, eventData.Event)
-		urd.LoadPos = eventData.LatestPos
-	}
+	urd.LoadPos = eventData.LatestPos
```
DMRobertson, line 503: Can we stick a comment above the declaration of LoadPos on the struct? (I keep forgetting what this means). Top of my head:

```
// LoadPos is an event NID. UserRoomData instances represent the status of this room after the corresponding event, as seen by this user.
```

sync3/handler/connstate.go
```diff
@@ -473,6 +473,8 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
 		var inviteState []json.RawMessage
 		// handle invites specially as we do not want to leak additional data beyond the invite_state and if
 		// we happen to have this room in the global cache we will do.
+		// Furthermore, rooms we have been invited to for the first time ever will not be in the global cache yet,
```
DMRobertson, line 476: who is "we"---the proxy as a whole, or the user in question? (I think the former?)

sync3/handler/connstate.go
```diff
@@ -473,6 +473,8 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
...
+		// which will cause errors below when we try calling functions on a nil metadata.
```
DMRobertson, line 477: Cross-linking: #85.

Off-topic here, but if you'll indulge me...

I can see that it's possible for the proxy to first learn of an room `R` is via an invite. But I don't understand what's special about invites here, as opposed to the other memberships. (Joins in particular being the most frequent, but also knocks. And what about kicks and bans?)

tests-e2e/security_test.go
```diff
@@ -117,15 +118,9 @@ func TestSecurityLiveStreamEventLeftLeak(t *testing.T) {
...
 			ID:     sensitiveEventID,
 		},
-	}, lastTwoEvents)
-	sensitiveEvent := lastTwoEvents[1]
-	kickEvent := lastTwoEvents[0]
-
-	// TODO: WE should be returning updated values for name and required_state
-	m.MatchResponse(t, aliceRes, m.MatchList("a", m.MatchV3Count(1)), m.MatchNoV3Ops(), m.MatchRoomSubscription(
-		roomID, m.MatchRoomTimelineMostRecent(2, []json.RawMessage{kickEvent, sensitiveEvent}),
-	))
```
DMRobertson, line 127: The point being here: we don't care if the events arrive in one sync response or two?

DMRobertson (approved):
To check: we expect this
- to spend more time in the DB  to provide sync responses, but
- to not randomly drop missed messages?

sync3/caches/user.go
kegsay: Bingo

tests-e2e/security_test.go
kegsay: Yes

sync3/handler/connstate.go
kegsay: The former, clarified.

sync3/handler/connstate.go
kegsay: Invites are special because they are the only membership state to occur prior to a join (excluding knocks which aren't implemented in any Element client, but have the same problems). Membership states which occur prior to a join are problematic for the proxy because there is no global state. The `invite_state` can change and get stale quickly, as it's only true at the point the invite was sent.

Join events will have global state associated with it. Leave/kick/bans all revert the state back to "not joined" which basically is just "remove this room from the list", i.e there is no room state to send to the client.

DMRobertson (approved):
Thanks!

PR #119, opened 2023-05-17
Migration: cleanup stale txns rows

DMRobertson:
And use a simpler query. What was I thinking???


sync2/tokens_table.go
```diff
@@ -137,10 +137,20 @@ type TokenForPoller struct {
...
-func (t *TokensTable) TokenForEachDevice() (tokens []TokenForPoller, err error) {
+// TokenForEachDevice loads the most recently used token for each device. If given a
+// transaction, it will SELECT inside that transaction.
+func (t *TokensTable) TokenForEachDevice(txn *sqlx.Tx) (tokens []TokenForPoller, err error) {
+	var db sqlx.Queryer
+	if txn != nil {
+		db = txn
+	} else {
+		db = t.db
+	}
```
DMRobertson, line 148: I originally wrote this as

```go
	var db sqlx.Queryer
	db = txn
	if db == nil {
		db = t.db
	}
```

But this falls afoul of the gotcha you pointed out to me some weeks ago: an interface with a prescribed type that holds a value of `nil` is not itself equal to nil. 

https://go.dev/tour/methods/12 and
https://glucn.medium.com/golang-an-interface-holding-a-nil-value-is-not-nil-bb151f472cc7 offer some context.

kegsay (approved):

PR #122, opened 2023-05-18
perf: improve startup speeds by using temp tables

kegsay:
When the proxy is run with large DBs (10m+ events), the startup queries are very slow (around 30min to load the initial snapshot.

After much EXPLAIN ANALYZEing, the cause is due to Postgres' query planner not making good decisions when the tables are that large. Specifically, the startup queries need to pull all joined members in all rooms, which ends up being nearly 50% of the entire events table of 10m rows. When this query is embedded in a subselect, the query planner assumes that the subselect will return only a few rows, and decides to pull those rows via an index. In this particular case, indexes are the wrong choice, as there are SO MANY rows a Seq Scan is often more appropriate. By using an index (which is a btree), this ends up doing log(n) operations _per row_ or `O(0.5 * n * log(n))` assuming we pull 50% of the table of n rows. As n increases, this is increasingly the wrong call over a basic O(n) seq scan. When n=10m, a seq scan has a cost of 10m, but using indexes has a cost of 16.6m. By dumping the result of the subselect to a temporary table, this allows the query planner to notice that using an index is the wrong thing to do, resulting in better performance. On large DBs, this decreases the startup time from 30m to ~5m.

On a backup database (which seems slower), this reduced startup queries from 1h35m to 6m24s, whilst keeping the end result byte for byte identical.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -120,17 +120,37 @@ func (s *Storage) InsertAccountData(userID, roomID string, events []json.RawMess
 	return data, err
 }
 
+// Prepare a snapshot of the database for calling snapshot functions.
+func (s *Storage) PrepareSnapshot(txn *sqlx.Tx) (tableName string, err error) {
+	// create a temporary table with all the membership nids for the current snapshots for all rooms.
+	// A temporary table will be deleted when the postgres session ends (this process quits).
+	// We insert these into a temporary table to let the query planner make better decisions. In practice,
+	// if we instead nest this SELECT as a subselect, we see very poor query times for large tables as
+	// each event NID is queried using a btree index, rather than doing a seq scan as this query will pull
+	// out ~50% of the rows in syncv3_events.
+	tempTableName := "temp_snapshot"
+	_, err = txn.Exec(
+		`SELECT UNNEST(membership_events) AS membership_nid INTO TEMP ` + tempTableName + ` FROM syncv3_snapshots
+		JOIN syncv3_rooms ON syncv3_snapshots.snapshot_id = syncv3_rooms.current_snapshot_id`,
+	)
+	return tempTableName, err
+}
+
 // GlobalSnapshot snapshots the entire database for the purposes of initialising
 // a sliding sync instance. It will atomically grab metadata for all rooms and all joined members
 // in a single transaction.
 func (s *Storage) GlobalSnapshot() (ss StartupSnapshot, err error) {
 	err = sqlutil.WithTransaction(s.accumulator.db, func(txn *sqlx.Tx) error {
-		var metadata map[string]internal.RoomMetadata
-		ss.AllJoinedMembers, metadata, err = s.AllJoinedMembers(txn)
+		tempTableName, err := s.PrepareSnapshot(txn)
 		if err != nil {
 			return err
 		}
-		err = s.MetadataForAllRooms(txn, metadata)
+		var metadata map[string]internal.RoomMetadata
+		ss.AllJoinedMembers, metadata, err = s.AllJoinedMembers(txn, tempTableName)
+		if err != nil {
+			return err
+		}
+		err = s.MetadataForAllRooms(txn, tempTableName, metadata)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


DMRobertson:
> the cause is due to Postgres' query planner not making good decisions when the tables are that large.

Something to bear in mind for the future is that there's some way of telling postgres "scale up/down your estimates". https://github.com/matrix-org/synapse/blob/07771fa487e1d281fdcae35a47db87ab675cb6b3/synapse/storage/schema/state/delta/61/02state_groups_state_n_distinct.sql.postgres#L29 is what I have in mind here.

kegsay:
> a) we just want it to always use index scans 

we want the opposite in this case.

DMRobertson:
> > a) we just want it to always use index scans
> 
> we want the opposite in this case.

Yes, but the point I'm getting at is

> -- By default the postgres statistics collector massively underestimates the
> -- number of distinct state groups are in the `state_groups_state`,

sounds a lot like:

> When this query is embedded in a subselect, the query planner assumes that the subselect will return only a few rows, [...]

Not saying we should do this Right Now; just mentioning for completeness.

state/storage.go
```diff
@@ -744,11 +756,10 @@ func (s *Storage) RoomMembershipDelta(roomID string, from, to int64, limit int)
...
 }
 
-func (s *Storage) AllJoinedMembers(txn *sqlx.Tx) (result map[string][]string, metadata map[string]internal.RoomMetadata, err error) {
+// Extract all rooms with joined members, and include the joined user list. Requires a prepared snapshot in order to be called.
+func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result map[string][]string, metadata map[string]internal.RoomMetadata, err error) {
 	rows, err := txn.Query(
-		`SELECT room_id, state_key from syncv3_events WHERE (membership='join' OR membership='_join') AND event_nid IN (
-			SELECT UNNEST(membership_events) FROM syncv3_snapshots JOIN syncv3_rooms ON syncv3_snapshots.snapshot_id = syncv3_rooms.current_snapshot_id
-		) ORDER BY event_nid ASC`,
+		`SELECT room_id, state_key from ` + tempTableName + ` INNER JOIN syncv3_events on membership_nid = event_nid WHERE membership='join' OR membership='_join' ORDER BY event_nid ASC`,
```
DMRobertson, line 762: Aside: why are we doing string concat for the table name here. Can you not pass table names with parameter placeholders?

state/storage.go
```diff
@@ -120,17 +120,37 @@ func (s *Storage) InsertAccountData(userID, roomID string, events []json.RawMess
...
+	// each event NID is queried using a btree index, rather than doing a seq scan as this query will pull
+	// out ~50% of the rows in syncv3_events.
+	tempTableName := "temp_snapshot"
+	_, err = txn.Exec(
+		`SELECT UNNEST(membership_events) AS membership_nid INTO TEMP ` + tempTableName + ` FROM syncv3_snapshots
+		JOIN syncv3_rooms ON syncv3_snapshots.snapshot_id = syncv3_rooms.current_snapshot_id`,
+	)
+	return tempTableName, err
+}
+
```
DMRobertson, line 138: Presumably if you call this twice then you'll end up duplicating all rows in the temp table?

DMRobertson: (Not that we ever do call it twice, of course)

state/storage.go
```diff
@@ -120,17 +120,37 @@ func (s *Storage) InsertAccountData(userID, roomID string, events []json.RawMess
 	return data, err
 }
 
+// Prepare a snapshot of the database for calling snapshot functions.
+func (s *Storage) PrepareSnapshot(txn *sqlx.Tx) (tableName string, err error) {
+	// create a temporary table with all the membership nids for the current snapshots for all rooms.
+	// A temporary table will be deleted when the postgres session ends (this process quits).
```
DMRobertson, line 126: It it really when the process ends? Naively I'd've thought it gets dropped when the transaction closes.

DMRobertson (approved):
> By using an index (which is a btree), this ends up doing log(n) operations per row or O(0.5 * n * log(n)) assuming we pull 50% of the table of n rows. As n increases, this is increasingly the wrong call over a basic O(n) seq scan.

So what, we now expect startup times to be linear in the number of events in the database?

state/storage.go
kegsay: Generally no, as table names are not params. Given the table name is a constant, this doesn't have an injection risk.

state/storage.go
kegsay: It would return an error actually as the table will already exist.

state/storage.go
kegsay: The reason why we don't use `ON COMMIT DROP` is because that syntax only works with `CREATE TABLE` which we aren't doing here.

kegsay:
Yes, or better.

Issue #78, opened 2023-04-18
Define and implement per-list bump_event_types


DMRobertson:
The only thing I'm not sure about is: what should we do if a room appears in two lists A and B which have different bump event types?

We already maintain per-conn per-room data; on the face of it, per-conn per-list per-room seems like a lot of effort. I'd be tempted to say:

> If a room appears in more than one list, it will be bumped in **all** lists whenever an event in **any** of the lists' `bump_event_types` are seen.
>
> For example, suppose  list A bumps rooms on `m.room.message` events and list B bumps rooms on `m.room.name` events. A room in list A and B will be bumped to the top of _both_ lists whenever a message occurs in the room **or** when the room's name changes.

However: we have informally discussed a mode of operation where the client requests multiple lists and sends an additional flag, asking that the sync responses merge these lists into one output list. (This would apparently be useful for mobile clients.) It might be worth bearing this in mind, as in this situation it's also possible that a room appears in multiple lists; we'd need to have well-defined semantics.

DMRobertson:
> on the face of it, per-conn per-list per-room seems like a lot of effort.

Kegan isn't convinced of this. The room conn metadata needs to track a map from list name to bump timestamp.

NB: any changes to bump_event_types won't take effect retroactively. Need to note this in the spec.

DMRobertson:
> However: we have informally discussed a mode of operation where ...

Pulling this out to a separate issue.

DMRobertson:
https://github.com/matrix-org/matrix-spec-proposals/commit/14580cae and https://github.com/matrix-org/matrix-spec-proposals/commit/4103ee768a4a3e1decee80c2987f50f4c6b3d539 define the new semantics.

PR #126, opened 2023-05-23
Distinguish global LastMessageTimestamp from conn LastActivityTimestamp

### LINKED ISSUE #46: refactor: metadata struct names
There's 3 tiers, all with bad names:
 - Global: `internal.RoomMetadata`
 - User `caches.UserRoomData`
 - Connection `sync3.RoomConnMetadata`

This isn't clear enough.

DMRobertson:
Progress towards https://github.com/matrix-org/sliding-sync/issues/78. I'm going to have a map from list name to timestamp, so first I want to have a place to put it.

Also try to leave comments that explain how the Dickens the metadata structs work.
Fixes #46.

### DIFF
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -109,8 +109,10 @@ func (s *ConnState) load(ctx context.Context) error {
 		rooms = append(rooms, sync3.RoomConnMetadata{
 			RoomMetadata: *metadata,
 			UserRoomData: urd,
-			// Best-effort only. Not going to be accurate if you've sent
-			// bump_event_types to exclude membership events.
+			// NB: If you've sent bump_event_types to exclude membership events and
+			// there are no interesting messages after your invite event, when you join
+			// this timestamp is going to roll back to the last interesting event before
+			// your invite.
 			LastInterestedEventTimestamp: metadata.LastMessageTimestamp,
 		})
 	}


sync3/handler/connstate.go
```diff
@@ -96,6 +96,10 @@ func (s *ConnState) load(ctx context.Context) error {
 		rooms[i] = sync3.RoomConnMetadata{
 			RoomMetadata: *metadata,
 			UserRoomData: urd,
+			// Best-effort only: we're not going to scan the database for all events in
+			// the entire room's history to give you a fully accurate timestamp
+			// according to your bump_event_types.
+			LastActivityTimestamp: metadata.LastMessageTimestamp,
```
DMRobertson, line 102: Maybe we should! See https://github.com/vector-im/element-x-ios/issues/413#issuecomment-1559836781

sync3/room.go
```diff
@@ -22,7 +22,19 @@ type Room struct {
 	NumLive           int               `json:"num_live,omitempty"`
 }
 
+// RoomConnMetadata represents a room as seen by one specific connection (hence once
+// specific device).
 type RoomConnMetadata struct {
+	// We enclose copies of the data kept in the global and user caches. These snapshots
+	// represent the state we reported to the connection the last time they requested
+	// a sync. Note that we are free to tweak fields within these copies if we want, to
+	// report more appropriate data to clients that kept in the caches.
```
DMRobertson, line 31: Wording is off here.

On tweaking fields: the only examples I could see of this today were

- the LastMessageTimestamp, which I explicitly want to pull out to a per-list map on the RoomConnMetadata; and
- the CanonicalisedName, which I'm suspicious of. I don't really understand why it needs to be computed in SetRoom.

sync3/caches/global.go
```diff
@@ -314,6 +314,8 @@ func (c *GlobalCache) OnNewEvent(
 			}
 		}
 	}
-	metadata.LastMessageTimestamp = ed.Timestamp
+	if ed.Timestamp > metadata.LastMessageTimestamp {
+		metadata.LastMessageTimestamp = ed.Timestamp
+	}
```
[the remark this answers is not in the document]
kegsay: this can be equally confusing if I see a message preview of 17:04 Hello world but it isn't sorted according to the 17:04 timestamp.

sync3/handler/connstate.go
```diff
@@ -105,6 +109,9 @@ func (s *ConnState) load(ctx context.Context) error {
 		rooms = append(rooms, sync3.RoomConnMetadata{
 			RoomMetadata: *metadata,
 			UserRoomData: urd,
+			// Best-effort only. Not going to be accurate if you've sent
+			// bump_event_types to exclude membership events.
```
kegsay, line 113: Why?

sync3/lists.go
```diff
@@ -74,16 +74,20 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
 			)
 		}
 
-		// Don't bump this room in the room list if the update isn't of interest to
-		// the client.
-		if !replacePreviousTimestamp {
-			r.LastMessageTimestamp = existing.LastMessageTimestamp
+		// Only bump this room in the room list if the update is of interest to the
```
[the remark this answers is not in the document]
kegsay: Unsure that is any clearer tbh. Let's see what the diff looks like in another PR. This metadata is shared across lists, so cloning it in this function may prove problematic.

sync3/room.go
```diff
@@ -22,7 +22,19 @@ type Room struct {
 	NumLive           int               `json:"num_live,omitempty"`
 }
 
+// RoomConnMetadata represents a room as seen by one specific connection (hence once
```
kegsay, line 25:
```suggestion
// RoomConnMetadata represents a room as seen by one specific connection (hence one
```

sync3/room.go
```diff
@@ -22,7 +22,19 @@ type Room struct {
...
+	// a sync. Note that we are free to tweak fields within these copies if we want, to
+	// report more appropriate data to clients that kept in the caches.
 	internal.RoomMetadata
 	caches.UserRoomData
+	// plus any per-conn data.
+
+	// LastActivityTimestamp is the largest origin_server_ts of events seen in the room
+	// that this connection is interested in. It should be less than or equal to
+	// internal.RoomMetadata.LastMessageTimestamp.
+	LastActivityTimestamp uint64
```
kegsay, line 39: You need to explain why this is in this struct and not RoomMetadata. i.e expand "that this connection is interested in" -> for bump event types.

sync3/sort.go
```diff
@@ -145,10 +145,10 @@ func (s *SortableRooms) comparatorSortByName(i, j int) int {
 
 func (s *SortableRooms) comparatorSortByRecency(i, j int) int {
 	ri, rj := s.resolveRooms(i, j)
-	if ri.LastMessageTimestamp == rj.LastMessageTimestamp {
+	if ri.LastActivityTimestamp == rj.LastActivityTimestamp {
 		return 0
 	}
-	if ri.LastMessageTimestamp > rj.LastMessageTimestamp {
+	if ri.LastActivityTimestamp > rj.LastActivityTimestamp {
```
kegsay, line 151: I wonder if `LastActivityTimestamp` is too vague here. It implies things like typing/receipts can affect the timestamp as they are activity. I understand not wanting to call it `LastMessageTimestamp` anymore though. Perhaps `LastInterestedEventTimestamp` or something? Slightly mouthy but it clarifies that it is indeed an `origin_server_ts` of an event, by some definition of "interested" which implies some kind of filter, without going too specific and mentioning bump event types.

kegsay (approved):

sync3/room.go
DMRobertson: I understand why it's computed eagerly; I don't understand why we compute it eagerly per-conn, rather than in the user cache itself.

sync3/sort.go
DMRobertson: Done in e665070.

sync3/room.go
DMRobertson: I.e. why would one conn have a different CanonicalisedName compared to any other for the same user, other than the conn being "behind" the latest data in the user cache?

sync3/caches/global.go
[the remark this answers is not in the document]
DMRobertson: I've backed this out in c953b39 following discussion.

sync3/handler/connstate.go
DMRobertson: This is a brand-new invite for a room we don't know about. But I guess in this situation there is literally no better timestamp available to use for sorting.

Hmm. Suppose that
- you exclude membership events from bump event types, but are interested in message events
- the last two events in the room are a message event (ts 50) and the invite (ts 100).

The room will originally appear in the list at ts 100 based on the invite data. After joining, the poller will get the room timeline and set the LastInterestedEventTimestamp to 50, pushing the room down the timeline. That sounds like it'd surprise me as a user.

DMRobertson: Will leave a comment noting this and move on.

PR #129, opened 2023-05-24
Per-list bump event types

### LINKED ISSUE #78: Define and implement per-list bump_event_types

DMRobertson:
Good enough for a first review pass.

Closes #78.

I still want to:
- [x] cleanup timestamp entries when a list gets deleted from the sync request
- [x] write an E2E test which demonstrates that the same room can be bumped independently in different lists

### DIFF
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -83,7 +83,7 @@ func NewConnState(
 //     N events arrive and get buffered.
 //   - load() bases its current state based on the latest position, which includes processing of these N events.
 //   - post load() we read N events, processing them a 2nd time.
-func (s *ConnState) load(ctx context.Context) error {
+func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
 	initialLoadPosition, joinedRooms, err := s.globalCache.LoadJoinedRooms(ctx, s.userID)
 	if err != nil {
 		return err
@@ -93,32 +93,40 @@ func (s *ConnState) load(ctx context.Context) error {
 	for _, metadata := range joinedRooms {
 		metadata.RemoveHero(s.userID)
 		urd := s.userCache.LoadRoomData(metadata.RoomID)
-		rooms[i] = sync3.RoomConnMetadata{
-			RoomMetadata: *metadata,
-			UserRoomData: urd,
+		timestamps := make(map[string]uint64, len(req.Lists))
+		for listKey, _ := range req.Lists {
 			// Best-effort only: we're not going to scan the database for all events in
 			// the entire room's history to give you a fully accurate timestamp
 			// according to your bump_event_types.
-			LastInterestedEventTimestamp: metadata.LastMessageTimestamp,
+			timestamps[listKey] = metadata.LastMessageTimestamp
+		}
+		rooms[i] = sync3.RoomConnMetadata{
+			RoomMetadata:                  *metadata,
+			UserRoomData:                  urd,
+			LastInterestedEventTimestamps: timestamps,
 		}
 		i++
 	}
 	invites := s.userCache.Invites()
 	for _, urd := range invites {
 		metadata := urd.Invite.RoomMetadata()
-		rooms = append(rooms, sync3.RoomConnMetadata{
-			RoomMetadata: *metadata,
-			UserRoomData: urd,
-			// NB: If you've sent bump_event_types to exclude membership events and
+		timestamps := make(map[string]uint64, len(req.Lists))
+		for listKey, _ := range req.Lists {
+			// NB: If you've set bump_event_types to exclude membership events and
 			// there are no interesting messages after your invite event, when you join
 			// this timestamp is going to roll back to the last interesting event before
 			// your invite.
-			LastInterestedEventTimestamp: metadata.LastMessageTimestamp,
+			timestamps[listKey] = metadata.LastMessageTimestamp
+		}
+		rooms = append(rooms, sync3.RoomConnMetadata{
+			RoomMetadata:                  *metadata,
+			UserRoomData:                  urd,
+			LastInterestedEventTimestamps: timestamps,
 		})
 	}
 
 	for _, r := range rooms {
-		s.lists.SetRoom(r, true)
+		s.lists.SetRoom(r)
 	}
 	s.loadPosition = initialLoadPosition
 	return nil
@@ -129,7 +137,7 @@ func (s *ConnState) OnIncomingRequest(ctx context.Context, cid sync3.ConnID, req
 	if s.loadPosition == -1 {
 		// load() needs no ctx so drop it
 		_, region := internal.StartSpan(ctx, "load")
-		s.load(ctx)
+		s.load(ctx, req)
 		region.End()
 	}
 	return s.onIncomingRequest(ctx, req, isInitial)
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -93,35 +93,35 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
 	for _, metadata := range joinedRooms {
 		metadata.RemoveHero(s.userID)
 		urd := s.userCache.LoadRoomData(metadata.RoomID)
-		timestamps := make(map[string]uint64, len(req.Lists))
+		interestedEventTimestampsByList := make(map[string]uint64, len(req.Lists))
 		for listKey, _ := range req.Lists {
 			// Best-effort only: we're not going to scan the database for all events in
 			// the entire room's history to give you a fully accurate timestamp
 			// according to your bump_event_types.
-			timestamps[listKey] = metadata.LastMessageTimestamp
+			interestedEventTimestampsByList[listKey] = metadata.LastMessageTimestamp
 		}
 		rooms[i] = sync3.RoomConnMetadata{
 			RoomMetadata:                  *metadata,
 			UserRoomData:                  urd,
-			LastInterestedEventTimestamps: timestamps,
+			LastInterestedEventTimestamps: interestedEventTimestampsByList,
 		}
 		i++
 	}
 	invites := s.userCache.Invites()
 	for _, urd := range invites {
 		metadata := urd.Invite.RoomMetadata()
-		timestamps := make(map[string]uint64, len(req.Lists))
+		inviteTimestampsByList := make(map[string]uint64, len(req.Lists))
 		for listKey, _ := range req.Lists {
 			// NB: If you've set bump_event_types to exclude membership events and
 			// there are no interesting messages after your invite event, when you join
 			// this timestamp is going to roll back to the last interesting event before
 			// your invite.
-			timestamps[listKey] = metadata.LastMessageTimestamp
+			inviteTimestampsByList[listKey] = metadata.LastMessageTimestamp
 		}
 		rooms = append(rooms, sync3.RoomConnMetadata{
 			RoomMetadata:                  *metadata,
 			UserRoomData:                  urd,
-			LastInterestedEventTimestamps: timestamps,
+			LastInterestedEventTimestamps: inviteTimestampsByList,
 		})
 	}
 
--- a/sync3/lists.go
+++ b/sync3/lists.go
@@ -55,7 +55,7 @@ func NewInternalRequestLists() *InternalRequestLists {
 	}
 }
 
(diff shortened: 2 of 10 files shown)


sync3/handler/connstate.go
```diff
@@ -93,32 +93,42 @@ func (s *ConnState) load(ctx context.Context) error {
...
-			RoomMetadata: *metadata,
-			UserRoomData: urd,
+		timestamps := make(map[string]uint64, len(req.Lists))
+		for listKey, _ := range req.Lists {
 			// Best-effort only: we're not going to scan the database for all events in
 			// the entire room's history to give you a fully accurate timestamp
 			// according to your bump_event_types.
-			LastInterestedEventTimestamp: metadata.LastMessageTimestamp,
+			timestamps[listKey] = metadata.LastMessageTimestamp
+		}
```
DMRobertson, line 102: This may not be unnecessary with the fallback logic I added for use in the sorter. But I think it's easier to be explicit as early as you can wherever possible.

sync3/handler/connstate.go
```diff
@@ -93,32 +93,40 @@ func (s *ConnState) load(ctx context.Context) error {
 	for _, metadata := range joinedRooms {
 		metadata.RemoveHero(s.userID)
 		urd := s.userCache.LoadRoomData(metadata.RoomID)
-		rooms[i] = sync3.RoomConnMetadata{
-			RoomMetadata: *metadata,
-			UserRoomData: urd,
+		timestamps := make(map[string]uint64, len(req.Lists))
```
DMRobertson, line 96: I tried using `s.muxedReq` here so that I wouldn't have to pass an argument through, but I got segfaults for my trouble. I guess this gets called before s.muxedReq is assigned a nonnil pointer.

sync3/lists.go
```diff
@@ -74,20 +74,32 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
...
 		}
 
-		// Only bump this room in the room list if the update is of interest to the
-		// client.
-		if replacePreviousTimestamp {
-			r.LastInterestedEventTimestamp = r.LastMessageTimestamp
-		} else {
-			r.LastInterestedEventTimestamp = existing.LastInterestedEventTimestamp
+		// Interpret the timestamp map on r as the changes we should apply atop the
+		// existing timestamps.
```
DMRobertson, line 78: I found it quite confusing to work out the best way to arrange this, and I'm not sure if this is the best way to do it. I think the end result is still a bit smelly.

In the else branch below, we trust that the caller has provided us with sensible timestamps. Doing it this way leaves the door open for us to feed in "accurate" initial timestamps, if we want to address https://github.com/vector-im/element-x-ios/issues/413#issuecomment-1559845357

sync3/lists.go
```diff
@@ -74,20 +74,32 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
...
+			newTs, bump := bumpListsTo[listKey]
+			if bump {
+				r.LastInterestedEventTimestamps[listKey] = newTs
+			} else {
+				prevTs, hadPreviousTs := existing.LastInterestedEventTimestamps[listKey]
+				if hadPreviousTs {
+					r.LastInterestedEventTimestamps[listKey] = prevTs
+				} else {
+					// This can happen if the listKey is brand-new in this request.
+					r.LastInterestedEventTimestamps[listKey] = existing.LastMessageTimestamp
```
DMRobertson, line 91: This is another best-effort affair.

sync3/lists.go
```diff
@@ -131,6 +143,7 @@ func (s *InternalRequestLists) RemoveRoom(roomID string) {
 
 func (s *InternalRequestLists) DeleteList(listKey string) {
 	// TODO
+	// TODO: could remove listKey's LastInterestedMessageTimestamp entries in allRooms
```
DMRobertson, line 146: I don't expect most clients will be repeatedly changing their list keys throughout the lifetime of one connection, but a malicious one could. That would eat up RAM. I think we probably ought to do this cleanup.

sync3/lists.go
```diff
@@ -153,8 +166,8 @@ func (s *InternalRequestLists) Get(listKey string) *FilteredSortableRooms {
 func (s *InternalRequestLists) ListsByVisibleRoomIDs(muxedReqLists map[string]RequestList) map[string][]string {
 	listsByRoomIDs := make(map[string][]string, len(muxedReqLists))
 	// Loop over each list, and mark each room in its sliding window as being visible in this list.
-	for listName, reqList := range muxedReqLists {
-		sortedRooms := s.lists[listName].SortableRooms
+	for listKey, reqList := range muxedReqLists {
```
DMRobertson, line 169: This isn't strictly relevant, but all the existing code uses `listKey`, apart from this bit that I wrote.

sync3/lists_test.go
```diff
@@ -41,7 +41,7 @@ func addRooms(list *sync3.InternalRequestLists, n int) {
 			UserRoomData: caches.UserRoomData{
 				IsDM: next%10 == 0,
 			},
-			LastInterestedEventTimestamp: messageTimestamp,
-		}, true)
+			LastInterestedEventTimestamps: make(map[string]uint64),
```
DMRobertson, line 44: Maybe I should build this properly to be the map from the keys of `list` to `messageTimestamp`? Can't remember why I didn't do so...

sync3/handler/connstate.go
kegsay: This needs a better name. I don't know what the keys are.

sync3/lists.go
```diff
@@ -74,20 +74,32 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
...
-		if replacePreviousTimestamp {
-			r.LastInterestedEventTimestamp = r.LastMessageTimestamp
-		} else {
-			r.LastInterestedEventTimestamp = existing.LastInterestedEventTimestamp
+		// Interpret the timestamp map on r as the changes we should apply atop the
+		// existing timestamps.
+		bumpListsTo := r.LastInterestedEventTimestamps
+		r.LastInterestedEventTimestamps = make(map[string]uint64, len(s.lists))
+		for listKey := range s.lists {
+			newTs, bump := bumpListsTo[listKey]
```
kegsay, line 82: `bumpListsTo` is a really poor name, and I wrote a comment before realising that this is the new map and not the existing map. Can we call it `newTimestamps` or something?

sync3/sort.go
```diff
@@ -145,10 +147,10 @@ func (s *SortableRooms) comparatorSortByName(i, j int) int {
 
 func (s *SortableRooms) comparatorSortByRecency(i, j int) int {
 	ri, rj := s.resolveRooms(i, j)
-	if ri.LastInterestedEventTimestamp == rj.LastInterestedEventTimestamp {
+	if ri.GetLastInterestedEventTimestamp(s.listKey) == rj.GetLastInterestedEventTimestamp(s.listKey) {
 		return 0
 	}
-	if ri.LastInterestedEventTimestamp > rj.LastInterestedEventTimestamp {
+	if ri.GetLastInterestedEventTimestamp(s.listKey) > rj.GetLastInterestedEventTimestamp(s.listKey) {
```
kegsay, line 153: It would be nice to precache this to avoid 2x map hits, i.e `tsRi := ri.GetLastInterestedEventTimestamp(s.listKey)` and `tsRj := rj.GetLastInterestedEventTimestamp(s.listKey)` then if `tsRi == rsRj` etc

sync3/room.go
```diff
@@ -33,16 +33,35 @@ type RoomConnMetadata struct {
...
 	// recency, this means rooms can suddenly fall down and jump back up the room
 	// list. See also the description of this in the React SDK docs:
 	//     https://github.com/matrix-org/matrix-react-sdk/blob/526645c79160ab1ad4b4c3845de27d51263a405e/docs/room-list-store.md#tag-sorting-algorithm-recent
-	LastInterestedEventTimestamp uint64
+	LastInterestedEventTimestamps map[string]uint64
+}
+
+func (r *RoomConnMetadata) GetLastInterestedEventTimestamp(listKey string) uint64 {
+	ts, ok := r.LastInterestedEventTimestamps[listKey]
+	if !ok {
```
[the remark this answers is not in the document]
DMRobertson: f5d5886

sync3/sort.go
DMRobertson: I was hoping that the compiler would be smart enough to do this for me. But I guess it has absolutely no idea what side effects the called function has. And even if it was just a plain lookup, it has no idea about what data races are going on behind the scenes.

sync3/lists.go
DMRobertson: My concern here isn't having one conn with thousands of simultaneous lists; it's a conn that

- creates list `a`
- next request, deletes list `a` and creates list `b`
- next request, deletes list `b` and creates list `c`
- next request, deletes list `c` and creates list `d`

etc

sync3/lists.go
```diff
@@ -74,20 +74,32 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
...
+				r.LastInterestedEventTimestamps[listKey] = newTs
+			} else {
+				prevTs, hadPreviousTs := existing.LastInterestedEventTimestamps[listKey]
+				if hadPreviousTs {
+					r.LastInterestedEventTimestamps[listKey] = prevTs
+				} else {
+					// This can happen if the listKey is brand-new in this request.
+					r.LastInterestedEventTimestamps[listKey] = existing.LastMessageTimestamp
+				}
+			}
```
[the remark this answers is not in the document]
DMRobertson: Yes, but note

> else
> use LastMessageTimestamp.

of _the existing_ RoomConnMetadata.

sync3/handler/connstate.go
DMRobertson: d8f158b has a better name.

sync3/lists_test.go
DMRobertson: Oh, I remember the problem. I can't get my hands on the list keys.

DMRobertson: 5f3c9df

sync3/lists.go
DMRobertson: I can do this, but again I think part of the confusion is that we are passing in an entirely new RoomConnMetadata `r` as an argument to this function. We really want to pass in a diff to an existing RoomConnMetadata.

DMRobertson: 11ba045

sync3/lists.go
```diff
@@ -74,20 +74,32 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata, replacePreviousTimest
...
 	} else {
 		// set the canonical name to allow room name sorting to work
 		r.CanonicalisedName = strings.ToLower(
 			strings.Trim(internal.CalculateRoomName(&r.RoomMetadata, 5), "#!():_@"),
 		)
-		// Also set a lastActivityTimestamp so recency sorting works.
-		r.LastInterestedEventTimestamp = r.LastMessageTimestamp
+		// We do not set LastInterestedEventTimestamps here. If there is no existing
+		// RoomConnMetadata, we use the timestamps provided by the caller. If those are
+		// missing, the sorter will we fall back to the LastMessageTimestamp.
```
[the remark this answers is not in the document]
DMRobertson: That's fair. I'll set this explicitly. 5333ed7

DMRobertson:
@kegsay: I have gone through your comments. Additionally, the last two commits are brand-new changes. PTAL when you are back.

sync3/lists.go
kegsay: Good point.

kegsay (approved):

PR #132, opened 2023-05-25
Fix an error in previous BumpEventTypes test

DMRobertson:
My error in thinking was noted in https://github.com/matrix-org/matrix-spec/issues/1545

### DIFF
--- a/tests-e2e/lists_test.go
+++ b/tests-e2e/lists_test.go
@@ -3,6 +3,7 @@ package syncv3_test
 import (
 	"fmt"
 	"github.com/matrix-org/sliding-sync/sync3/extensions"
+	"github.com/tidwall/gjson"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


tests-e2e/lists_test.go
```diff
@@ -994,7 +1001,7 @@ func TestBumpEventTypesHandling(t *testing.T) {
 				t.Fatalf("expected no ops while waiting for receipt: %s", err)
 			}
 			matchReceipt := m.MatchReceipts(room2, []m.Receipt{{
-				EventID: charlieJoinEventID,
+				EventID: aliceLastSeenEventID,
```
DMRobertson, line 1004: I was kinda surprised this worked before. But I guess the receipt event IDs are opaque to the homeserver?

kegsay: I guess so! :S

PR #134, opened 2023-05-25
Associate user IDs and device IDs to poller sentry reports

DMRobertson:
I noticed that this might be helpful when looking through Sentry errors.

See the commit message of e5eb4f12bafb960a3389accba4ef22fb57721cb2  for more rationale.

Completely untested.

### DIFF
--- a/internal/sentry.go
+++ b/internal/sentry.go
@@ -6,6 +6,10 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


kegsay (approved):

DMRobertson:
I manually checked this by running an E2E test with pollers altered to log an event to sentry every iteration of the polling loop. I could see a username and device ID logged to sentry alongside the traceback.

Issue #38, opened 2023-03-22
Add support for MSC3946: Dynamic room predecessor


andybalaam:
[MSC3946](https://github.com/matrix-org/matrix-spec-proposals/pull/3946) enables us to specify the predecessor room using a new state event instead of the create event.

This doesn't have much effect on the sliding sync proxy, but it does affect the list of rooms that are returned under certain circumstances.

As I understand it, predecessor rooms should normally be excluded from the list of returned rooms, so the proxy will need to determine the predecessor using the new state event if present. I believe this will be needed if the `include_old_rooms` filter is active.

I implemented this stuff in matrix-js-sdk - happy to help if you have any questions.  The most relevant code is probably in `findPredecessor` in [room-state.ts](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/models/room-state.ts#L981).

DMRobertson:
We do similar things for m.room.create as part of supporting the `include_old_rooms` flag. It should be easy to replicate the lookup.

The problem is that we had assumed that predecessors never change. But now it can. Will have to update some SQL and also the caches.

Relevant source:
https://github.com/matrix-org/sliding-sync/blob/main/state/rooms_table.go#L15
https://github.com/matrix-org/sliding-sync/blob/main/state/rooms_table.go#L49 need to be more selective about what "latches on"

PR #139, opened 2023-05-31
WIP: Support for msc3946

DMRobertson:
Closes https://github.com/matrix-org/sliding-sync/issues/38

Utterly WIP. I'm opening the PR for better visibility of in-flight work.

Concerns:

- What is there to stop me from marking rooms A and B as both superseding room C? At present the latest seen predecessor event wins.
- If I mark room A as superseding B1, then mark A as superseding B2, is B1 still considered to be replaced by A?
- What do we do if we learn that A supersedes B but we have no knowledge of room B? It would be a shame if we later joined B and didn't present it as being replaced by B.

### DIFF
--- a/pubsub/v2.go
+++ b/pubsub/v2.go
@@ -36,6 +36,10 @@ type V2Accumulate struct {
 	RoomID    string
 	PrevBatch string
 	EventNIDs []int64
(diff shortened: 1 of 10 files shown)


DMRobertson:
Going to drop this. I didn't get far, it's not a priority, and there looks to have been some bitrot.

PR #136, opened 2023-05-26
Reproduce and avoid panic in RemoveHero

DMRobertson:
Intended to fix #85.

Commitwise reviewable.

PR #140, opened 2023-06-01
WIP make bump event types work on startup

DMRobertson:


state/storage.go
```diff
@@ -182,13 +182,22 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
...
 	if err != nil {
 		return err
 	}
 	for _, ev := range events {
-		metadata := result[ev.RoomID]
+		metadata, ok := result[ev.RoomID]
 		metadata.LastMessageTimestamp = gjson.ParseBytes(ev.JSON).Get("origin_server_ts").Uint()
+		if !ok {
+			metadata = *internal.NewRoomMetadata(ev.RoomID)
+		}
```
DMRobertson, line 194: I'm not sure if this is needed. The comment below makes it sound like the metadata may not be present in the map.  If so, I don't want to write to a `LatestEventsByType` which is a nil map, because AFAIK that panics.

If that reasoning is sound, why did we get away with this before? internal.RoomMetadata has a ChildSpaceRooms map which is nil unless you use the NewRoomMetadata helper?

sync3/caches/global.go
```diff
@@ -101,23 +102,31 @@ func (c *GlobalCache) LoadRooms(ctx context.Context, roomIDs ...string) map[stri
 	return result
 }
 
-// Load all current joined room metadata for the user given. Returns the absolute database position along
-// with the results. TODO: remove with LoadRoomState?
-func (c *GlobalCache) LoadJoinedRooms(ctx context.Context, userID string) (pos int64, joinedRooms map[string]*internal.RoomMetadata, err error) {
+// Load all current joined room metadata for the user given, augmented with the NID of
```
DMRobertson, line 105: `augmented with` needs rewording.

state/storage.go
```diff
@@ -785,38 +794,53 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result m
...
+			delete(joinedNIDsByRoom, ev.RoomID)
 		}
 	}
-	joinedRooms := make([]string, 0, len(joinedRoomsSet))
-	for roomID, joined := range joinedRoomsSet {
-		if joined {
-			joinedRooms = append(joinedRooms, roomID)
-		}
+	joinedRoomIDs = make([]string, 0, len(joinedNIDsByRoom))
+	joinedRoomNIDs = make([]int64, 0, len(joinedNIDsByRoom))
```
DMRobertson, line 837: I suppose the `0` here is unnecessary.

sync3/caches/user.go
```diff
@@ -210,10 +212,14 @@ func (c *UserCache) Unsubscribe(id int) {
 	delete(c.listeners, id)
 }
 
+// OnRegistered is called after the sync3.Dispatcher has successfully registered this
+// cache to receive updates. We use this to run some final initialisation logic that
+// is sensitive to race conditions; confusingly, most of the initialisation is driven
+// externally by sync3.SyncLiveHandler.userCache.
```
DMRobertson, line 218: I think this is accurate(?)

sync3/caches/user.go
```diff
@@ -263,6 +270,17 @@ func (c *UserCache) OnRegistered(ctx context.Context, _ int64) error {
...
+		// Record when we joined the room. We've just had to scan the history of our
+		// membership in this room to produce joinedRooms above, so we may as well
+		// do this here too.
+		c.roomToDataMu.Lock()
+		urd, ok := c.roomToData[room.RoomID]
+		if !ok {
+			urd = NewUserRoomData()
+		}
+		urd.JoinNID = joinNIDs[room.RoomID]
+		c.roomToDataMu.Unlock()
```
DMRobertson, line 283: Would it make more sense to have a second loop whose only job is to write `JoinNID`, and to take the lock outside of the loop?

sync3/dispatcher.go
```diff
@@ -19,10 +19,14 @@ var logger = zerolog.New(os.Stdout).With().Timestamp().Logger().Output(zerolog.C
 
 const DispatcherAllUsers = "-"
 
+// Receiver represents the callbacks that a Dispatcher may fire.
 type Receiver interface {
 	OnNewEvent(ctx context.Context, event *caches.EventData)
 	OnReceipt(ctx context.Context, receipt internal.Receipt)
 	OnEphemeralEvent(ctx context.Context, roomID string, ephEvent json.RawMessage)
+	// OnRegistered is called after a successful call to Dispatcher.Register. After
+	// this call, the receiver will be told about events whose NID is greater than
```
DMRobertson, line 28: s/events/PDUs

DMRobertson: (i.e. the receiver always gets told about all EDUs; EDUs have nothing like a NID to act as a sequence number)

sync3/dispatcher.go
```diff
@@ -31,7 +35,8 @@ type Dispatcher struct {
 	jrt              *JoinedRoomsTracker
 	userToReceiver   map[string]Receiver
 	userToReceiverMu *sync.RWMutex
-	latestPos        int64
+	// latestPos is an eventNID, the largest that this dispatcher has seen.
+	latestPos int64
```
DMRobertson, line 39: Aside: a comment on WTF the dispatcher does would be helpful. It sounds like its job is to post-process certain V2Listener messages and re-publish them in a form that is suitable for the caches?

sync3/handler/connstate.go
```diff
@@ -93,6 +93,9 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
 	for _, metadata := range joinedRooms {
 		metadata.RemoveHero(s.userID)
 		urd := s.userCache.LoadRoomData(metadata.RoomID)
+		// TODO: is this necessary---will the joinNID already be in the UserCache?
+		urd.JoinNID = joinNIDs[metadata.RoomID]
+
```
DMRobertson, line 98: Not sure on this one.

sync3/handler/handler.go
```diff
@@ -440,6 +440,14 @@ func (h *SyncLiveHandler) CacheForUser(userID string) *caches.UserCache {
...
+// it
+//  - creates a blank caches.UserCache struct,
+//  - fires callbacks on that struct as necessary to populate it with initial state,
+//  - stores the struct so it will not be recreated in the future, and
+//  - registers the cache with the Dispatcher.
+//
+// Some extra initialisation takes place in caches.UserCache.OnRegister.
 func (h *SyncLiveHandler) userCache(userID string) (*caches.UserCache, error) {
 	// bail if we already have a cache
 	c, ok := h.userCaches.Load(userID)
```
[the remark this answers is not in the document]
kegsay: This can be refactored away. It was originally like this because the user cache did not have a `storage` to work with, i.e all input data was passed into it, which made testing the cache easier. Nowadays we do need the `storage` layer, so the distinction isn't worth keeping.

state/storage.go
```diff
@@ -785,38 +794,53 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result m
 	return result, metadata, nil
 }
 
-func (s *Storage) JoinedRoomsAfterPosition(userID string, pos int64) ([]string, error) {
+func (s *Storage) JoinedRoomsAfterPosition(userID string, pos int64) (
+	joinedRoomIDs []string, joinedRoomNIDs []int64, err error,
```
kegsay, line 798: joined _room_ nids? This is confusing, surely you mean `joinEventNIDs`?

sync3/handler/connstate.go
kegsay: No, it won't be. `userCache.LoadRoomData` _lazily_ loads data, which is seeded from `(h *SyncLiveHandler) userCache(userID string)`. As you don't seed the join nids in that function, it will be unset. Seeding happens by calling `userCache.On....` callbacks from data from the database in order to reuse codepaths.

By setting this here, this will not take effect in the user cache (`LoadRoomData` returns copies of metadata precisely to avoid callers polluting the cache). To get this to stick around, you need to add a new function to `UserCache` which will assign the join nids and then call that in the handler to seed the cache correctly. It's a bit bleh, since the user cache doesn't know the list of joined rooms at that point in time, so you'll need to hit `s.globalCache.LoadJoinedRooms(ctx, s.userID)` twice for new conns (once to seed the cache, once for the conn).


EDIT: I wrote all of this, and then read the code in `OnRegistered`, which does exactly this (`LoadJoinedRooms` then sets join nids). So ignore the above.

state/storage.go
```diff
@@ -785,38 +794,53 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result m
...
-		// end result, not the deltas
 		membership := gjson.GetBytes(ev.JSON, "content.membership").Str
 		switch membership {
+		// These are "join" and the only memberships that you can transition to after
+		// a join: see e.g. the transition diagram in
+		// https://spec.matrix.org/v1.7/client-server-api/#room-membership
 		case "join":
-			joinedRoomsSet[ev.RoomID] = true
+			if _, alreadyJoined := joinedNIDsByRoom[ev.RoomID]; !alreadyJoined {
+				joinedNIDsByRoom[ev.RoomID] = ev.NID
```
kegsay, line 828: This is slightly unclear as it gives the impression the first join event we see will be the NID. Only once you see that we delete from `joinedNIDsByRoom` on leaves/bans does it make sense that `!alreadyJoined` will be true again.

state/storage.go
```diff
@@ -785,38 +794,53 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result m
...
-	for roomID, joined := range joinedRoomsSet {
-		if joined {
-			joinedRooms = append(joinedRooms, roomID)
-		}
+	joinedRoomIDs = make([]string, 0, len(joinedNIDsByRoom))
+	joinedRoomNIDs = make([]int64, 0, len(joinedNIDsByRoom))
+	for roomID, nid := range joinedNIDsByRoom {
+		joinedRoomIDs = append(joinedRoomIDs, roomID)
+		joinedRoomNIDs = append(joinedRoomNIDs, nid)
 	}
```
kegsay, line 841: It seems a shame to unzip the nice map of room ID to event NID into 2 slices :(

sync3/dispatcher.go
kegsay: Yeah, as per previous convos, the dispatcher can be folded into the Handler at this point.

sync3/caches/user.go
kegsay: Yes, `OnRegistered` and spaces handling came much later in development hence the weird multiple places. This _wasn't_ folded into one function because `OnRegistered` has taken out a global lock on the dispatcher so it's important that we don't spend too long inside this function.

sync3/caches/user.go
kegsay: This is fine for now, even though it will be locking N times (once per room). We'd see it in the profiler if it's causing latency problems.

sync3/caches/user.go
```diff
@@ -263,6 +270,17 @@ func (c *UserCache) OnRegistered(ctx context.Context, _ int64) error {
...
+
+		// Record when we joined the room. We've just had to scan the history of our
+		// membership in this room to produce joinedRooms above, so we may as well
+		// do this here too.
+		c.roomToDataMu.Lock()
+		urd, ok := c.roomToData[room.RoomID]
+		if !ok {
+			urd = NewUserRoomData()
+		}
+		urd.JoinNID = joinNIDs[room.RoomID]
```
kegsay, line 282: You need to remember to set the value, else it won't take effect as you aren't dealing with pointers!
```suggestion
		urd.JoinNID = joinNIDs[room.RoomID]
		c.roomToData[room.RoomID] = urd
```

sync3/caches/global.go
```diff
@@ -101,23 +102,31 @@ func (c *GlobalCache) LoadRooms(ctx context.Context, roomIDs ...string) map[stri
...
 	}
-	joinedRoomIDs, err := c.store.JoinedRoomsAfterPosition(userID, initialLoadPosition)
+	joinedRoomIDs, joinNIDsSlice, err := c.store.JoinedRoomsAfterPosition(userID, initialLoadPosition)
 	if err != nil {
-		return 0, nil, err
+		return 0, nil, nil, err
+	}
+	joinNIDs = make(map[string]int64, len(joinNIDsSlice))
+	for i, nid := range joinNIDsSlice {
+		joinNIDs[joinedRoomIDs[i]] = nid
```
kegsay, line 125: Amusing that we immediately zip up the two slices.. can we just make `JoinedRoomsAfterPosition` return `map[string]int64`?

kegsay (approved):

state/storage.go
DMRobertson: I think I was confusing direct insertion versus appending.

(I still think it's really confusing that `len(make([]V, 10)) == 10` but `len(make(map[K]V, 10)) == 0`.

state/storage.go
DMRobertson: Will use `currentlyJoined` and add a comment.

sync3/caches/global.go
DMRobertson: b3544a5 attempts this, PTAL

state/storage.go
```diff
@@ -182,13 +182,22 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
...
 	// work out latest timestamps
-	events, err := s.accumulator.eventsTable.selectLatestEventInAllRooms(txn)
+	events, err := s.accumulator.eventsTable.selectLatestEventByTypeInAllRooms(txn)
 	if err != nil {
 		return err
 	}
 	for _, ev := range events {
-		metadata := result[ev.RoomID]
+		metadata, ok := result[ev.RoomID]
 		metadata.LastMessageTimestamp = gjson.ParseBytes(ev.JSON).Get("origin_server_ts").Uint()
```
kegsay, line 191: This will be wrong now surely as we are just constantly setting this, so unless `selectLatestEventByTypeInAllRooms` orders by origin_server_ts, it will be set to a random timestamp for one of the event types. We need to use the max.

sync3/caches/global.go
```diff
@@ -76,49 +77,73 @@ func (c *GlobalCache) OnRegistered(_ context.Context, _ int64) error {
...
+
+// copyRoom returns a copy of the internal.RoomMetadata stored for this room.
+// This is an internal implementation detail of LoadRooms and LoadRoomsFromMap.
+// If the room is not present in the global cache, returns a stub metadata entry.
+// The caller MUST acquire a read lock on roomIDToMetadataMu before calling this.
+func (c *GlobalCache) copyRoom(roomID string) *internal.RoomMetadata {
+	sr := c.roomIDToMetadata[roomID]
+	if sr == nil {
+		logger.Warn().Str("room", roomID).Msg("GlobalCache.LoadRoom: no metadata for this room, generating stub")
+		c.roomIDToMetadata[roomID] = internal.NewRoomMetadata(roomID)
```
kegsay, line 113: Comments on this function suggest a read lock is needed, but here we write.

kegsay (approved):

sync3/handler/connstate.go
```diff
@@ -93,6 +93,9 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
 	for _, metadata := range joinedRooms {
 		metadata.RemoveHero(s.userID)
 		urd := s.userCache.LoadRoomData(metadata.RoomID)
+		// TODO: is this necessary---will the joinNID already be in the UserCache?
```
DMRobertson, line 96:
```suggestion
```

Per https://github.com/matrix-org/sliding-sync/pull/140/files#r1214570794

sync3/caches/global.go
DMRobertson: Oh true. An error in https://github.com/matrix-org/sliding-sync/pull/136 then?

DMRobertson: Thought process:
- could acquire a write lock in the caller, but I'm guessing LoadRooms is called frequently, and don't want to nobble concurrent reads
- maybe we just return a blank stub but don't insert it into the map?

PR #141, opened 2023-06-02
perf: reduce the time taken to pull out your own receipts

kegsay:
By doing a single bulk query rather than one-per-room.

### DIFF
--- a/state/receipt_table.go
+++ b/state/receipt_table.go
@@ -97,20 +97,30 @@ func (t *ReceiptTable) SelectReceiptsForEvents(roomID string, eventIDs []string)
 }
 
 // Select all (including private) receipts for this user in this room.
-func (t *ReceiptTable) SelectReceiptsForUser(roomID, userID string) (receipts []internal.Receipt, err error) {
+func (t *ReceiptTable) SelectReceiptsForUser(roomIDs []string, userID string) (receiptsByRoom map[string][]internal.Receipt, err error) {
+	var receipts []internal.Receipt
 	err = t.db.Select(&receipts, `SELECT room_id, event_id, user_id, ts, thread_id FROM syncv3_receipts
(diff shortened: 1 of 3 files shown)


state/receipt_table.go
```diff
@@ -97,20 +97,30 @@ func (t *ReceiptTable) SelectReceiptsForEvents(roomID string, eventIDs []string)
 }
 
 // Select all (including private) receipts for this user in this room.
```
DMRobertson, line 99:
```suggestion
// Select all (including private) receipts for this user in these rooms.
```

sync3/extensions/receipts.go
```diff
@@ -95,18 +98,29 @@ func (r *ReceiptsRequest) ProcessInitial(ctx context.Context, res *Response, ext
...
+	for roomID, ownRecs := range ownReceipts {
+		otherReceipts[roomID] = append(otherReceipts[roomID], ownRecs...)
+	}
+
+	for roomID, receipts := range otherReceipts {
+		if len(receipts) == 0 {
 			continue
 		}
-		rooms[roomID], _ = state.PackReceiptsIntoEDU(append(receipts, ownReceipts...))
+		rooms[roomID], _ = state.PackReceiptsIntoEDU(append(receipts, receipts...))
```
DMRobertson, line 121: This doesn't look right. Should this be `state.PackReceiptsIntoEDU(receipts)`?

kegsay: Yes, I forgot this bit. This would've duplicated receipts which is why tests were still happy.

kegsay:
```suggestion
		rooms[roomID], _ = state.PackReceiptsIntoEDU(receipts)
```

DMRobertson (approved):

PR #142, opened 2023-06-06
Make bump event types work on startup

DMRobertson:


state/storage.go
```diff
@@ -188,10 +188,15 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
...
-		metadata.LastMessageTimestamp = gjson.ParseBytes(ev.JSON).Get("origin_server_ts").Uint()
 		if !ok {
 			metadata = *internal.NewRoomMetadata(ev.RoomID)
 		}
+		// For a given room, we'll see many events (one for each event type in the
+		// room's state). We need to pick the largest of these events' timestamps here.
+		ts := gjson.ParseBytes(ev.JSON).Get("origin_server_ts").Uint()
+		if ts > metadata.LastMessageTimestamp {
+			metadata.LastMessageTimestamp = ts
+		}
```
DMRobertson, line 199: You pointed this out in https://github.com/matrix-org/sliding-sync/pull/140/files#r1219292296 and I forgot to commit this change before merging.

Additionally, the write to LastMessageTimestamp needs to happen _after_ the call to NewRoomMetadata.

sync3/caches/global.go
```diff
@@ -31,10 +31,12 @@ type EventData struct {
...
 
-	// the absolute latest position for this event data. The NID for this event is guaranteed to
-	// be <= this value. See PosAlwaysProcess and PosDoNotProcess for things outside the event timeline
-	// e.g invites
-	LatestPos int64
+	// NID is the nid for this event; or a non-nid sentinel value. Current sentinels are
+	//  - PosAlwaysProcess and PosDoNotProcess, for things outside the event timeline
+	//    e.g invites; and
+	//  - `0` used by UserCache.OnRegistered to inject space children events at startup.
+	//    It's referenced in ConnState.OnRoomUpdateand UserCache.OnSpaceUpdate
```
DMRobertson, line 38: I'm a little uneasy about these sentinels, to be honest.

tests-integration/lists_test.go
```diff
@@ -191,3 +191,167 @@ func TestUnreadCountMisordering(t *testing.T) {
...
+	ts = ts.Add(time.Second)
+	r1Timeline = append(r1Timeline,
+		testutils.NewStateEvent(
+			t, "m.room.topic", "", alice, map[string]interface{}{"topic": "potato"}, testutils.WithTimestamp(ts),
+		),
+	)
+
+	ts = ts.Add(time.Second)
+	r1Timeline = append(r1Timeline,
+		testutils.NewMessageEvent(t, alice, "message in room 1", testutils.WithTimestamp(ts)),
```
kegsay, line 229: I would find this easier to read if this was table-driven and we added a helper to set the timestamp after-the-fact i.e:
```go
ts := time.Now()
steps := []struct{
    timeline *[]json.RawMessage
    event json.RawMessage
}{
    {
        timeline: &r1Timeline,
        event: testutils.NewMessageEvent(t, alice, "message in room 1"),
    },
    // ...
}

for _, step := range steps {
     ts = ts.Add(time.Second)
     step.event = setTimstamp(step.event, ts)
     step.timeline = append(step.timeline, step.event)
}
```

The usage of `*[]json.RawMessage` means you can use `r1Timeline` and have it refer to the new events correctly.

tests-integration/lists_test.go
```diff
@@ -191,3 +191,167 @@ func TestUnreadCountMisordering(t *testing.T) {
...
+	r3Timeline := []json.RawMessage{}
+
+	t.Log("A series of events take place in all three rooms.")
+	// The following sequence of events take place:
+	// (1) r1: topic set
+	// (2) r1: message
+	// (3) r2: message
+	// (4) r3: message
+	// (5) r2: topic
+	// (6) r1: profile change
```
kegsay, line 218: This should be by the test cases for clarity.

tests-integration/lists_test.go
```diff
@@ -191,3 +191,167 @@ func TestUnreadCountMisordering(t *testing.T) {
...
+		BumpEventTypes []string
+		RoomIDs        []string
+	}{
+		{
+			BumpEventTypes: []string{"m.room.message"},
+			RoomIDs:        []string{room3ID, room2ID, room1ID},
+		},
+		{
+			BumpEventTypes: []string{"m.room.topic"},
+			RoomIDs:        []string{room2ID, room1ID, room3ID},
```
kegsay, line 321: Would be nice to xref values in the block comment above to justify the values you give here. I've checked them all but it would help if you could help the reader more. i.e:
```go
BumpEventTypes: []string{"m.room.topic"},
// r2 topic (5) then r1 topic (1) then no topic
RoomIDs:        []string{room2ID, room1ID, room3ID},
```

tests-e2e/lists_test.go
```diff
@@ -1116,6 +1116,129 @@ func TestBumpEventTypesInOverlappingLists(t *testing.T) {
...
+	t.Log("Bob should see room1 ahead of room2.")
+	// The order in which things happen:
+	// 1. alice: room1 secret event
+	// 2. bob:   room2 secret event
+	// 3. alice: invite bob to room 1
+	// 4. bob:   join room 1
+	// Bob can only see (2), (3) and (4), which means that room 1 has had most recent activity.
+	// (If we use the secret events' timestamps alone, without considering what Bob has
+	// permission to see, we will only consider (1) and (2), which would mean room 2
+	// has had the most recent activity.)
```
kegsay, line 1169: Devious! Great test.

sync3/request.go
```diff
@@ -260,6 +260,11 @@ func (rl *RequestList) WriteSwapOp(
 	}
 }
 
+// HasBumpEventTypes returns true iff this request has specified bump_event_types.
+func (rl *RequestList) HasBumpEventTypes() bool {
+	return rl.BumpEventTypes != nil && len(rl.BumpEventTypes) > 0
```
kegsay, line 265: The nil check is superfluous here, as a nil slice has length 0.

sync3/caches/user.go
```diff
@@ -530,7 +531,7 @@ func (c *UserCache) OnSpaceUpdate(ctx context.Context, parentRoomID, childRoomID
 func (c *UserCache) OnNewEvent(ctx context.Context, eventData *EventData) {
 	// add this to our tracked timelines if we have one
 	urd := c.LoadRoomData(eventData.RoomID)
-	urd.LoadPos = eventData.LatestPos
+	urd.LoadPos = eventData.NID
```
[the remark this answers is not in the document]
kegsay: Sure why not.

sync3/caches/user.go
```diff
@@ -281,7 +282,7 @@ func (c *UserCache) OnRegistered(ctx context.Context, _ int64) error {
 		if !ok {
 			urd = NewUserRoomData()
 		}
-		urd.JoinNID = joinNIDs[room.RoomID]
+		urd.JoinTiming = joinTimings[room.RoomID]
```
kegsay, line 285: Is it possible for there to be no join timings and hence `urd.JoinTiming == nil`? Should init an empty map then?

sync3/handler/connstate.go
```diff
@@ -93,18 +93,29 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
...
+			interestingActivityTs := metadata.LastMessageTimestamp
+			if listReq.HasBumpEventTypes() {
+				// Use the global cache to find the timestamp of the latest interesting
+				// event we can see. If there is no such event, fall back to the
+				// LastMessageTimestamp.
+				joinEvent := joinTimings[metadata.RoomID]
+				interestingActivityTs = joinEvent.Timestamp
+				for _, eventType := range listReq.BumpEventTypes {
+					timing := metadata.LatestEventsByType[eventType]
+					if joinEvent.NID < timing.NID && interestingActivityTs < timing.Timestamp {
```
kegsay, line 109:
```suggestion
					// we found a later event which we are authorised to see, use it instead
					if joinEvent.NID < timing.NID && interestingActivityTs < timing.Timestamp {
```

sync3/lists.go
```diff
@@ -97,17 +97,8 @@ func (s *InternalRequestLists) SetRoom(r RoomConnMetadata) (delta RoomDelta) {
...
 		)
-		// Set LastInterestedEventTimestamps here so that recency sorts work. We use the
-		// timestamps provided by the caller, but otherwise fall back to that in the
-		// global event metadata.
-		newTimestamps := r.LastInterestedEventTimestamps
-		r.LastInterestedEventTimestamps = make(map[string]uint64, len(s.lists))
-		for listKey := range s.lists {
-			_, ok := newTimestamps[listKey]
-			if !ok {
-				newTimestamps[listKey] = r.LastMessageTimestamp
```
kegsay, line 108: I don't know why we're removing this. We haven't changed how the caller behaves, which is:
 - make a new map: `bumpTimestampInList := make(map[string]uint64, len(s.muxedReq.Lists))`
 - if there is a room update:
    * the list isn't using bump event types: set latest timestamp
    * the list is using bump event types: set only if event matches
 - call SetRoom with this map.

This code being deleted was useful as it preserved the previous values if they were not modified, whereas now they are being forgotten?

kegsay (approved):

tests-integration/lists_test.go
```diff
@@ -191,3 +191,167 @@ func TestUnreadCountMisordering(t *testing.T) {
...
+		roomID: room2ID,
+		name:   "room 2",
+		state:  r2CreateState,
+		events: r2Timeline,
+	}
+	const room3ID = "!room3:localhost"
+	r3 := roomEvents{
+		roomID: room3ID,
+		name:   "room 3",
+		state:  r3CreateState,
```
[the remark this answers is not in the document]
DMRobertson: I think it was just because I wanted to spell out the sequence events in chronological order.

sync3/request.go
DMRobertson: Oh true enough. /me inlines

sync3/caches/user.go
DMRobertson: Good question. For this to be the case, the `joinedRooms` and `joinTimings` returned by `GlobalCache.LoadJoinedRooms` would have to return maps with different sets of keys. I would hope this is impossible---will check and leave a comment if so.

DMRobertson: Oh, I misunderstood you. You're talking about the entire `joinTimings` map being nil.

DMRobertson: > Is it possible for there to be no join timings and hence urd.JoinTiming == nil?

No, I've traced through the logic and concluded that LoadJoinedRooms either returns `(nil, nil, err)` or `(x, y, nil)` where x and y are non-nil maps with the same set of keys. I've left comments asserting this.

sync3/lists.go
DMRobertson: AFAICS, The code that was deleted is plain wrong.

This takes place in the `else` branch (see -95 and -60) which is only called when we don't have a previous, already `existing` `*RoomConnMetadata`. I spy two wrongnesses:

1. `r.LastInterestedEventTimestamps` is set to an empty map on -104, but then _never written to afterwards_. 
2. `s.lists` is empty before `AssignList` is called for the first time. I believe the first `SetRoom` call happens before this(?)---if so, the loop on -105 does nothing.

By my reading, the old code completely dropped the LastInterestedTimestamps provided by the first SetRoom call.

PR #144, opened 2023-06-06
Fix a panic on startup with outstanding invite

DMRobertson:
if the proxy is not aware of anyone else joined to the room.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -209,6 +209,8 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
 		metadata.LatestEventsByType[parsed.Get("type").Str] = eventMetadata
 		// it's possible the latest event is a brand new room not caught by the first SELECT for joined
 		// rooms e.g when you're invited to a room so we need to make sure to set the metadata again here
+		// TODO: is this accurate now that we explicitly call NewRoomMetadata above
+		//       when handling invites?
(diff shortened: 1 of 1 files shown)


state/storage.go
```diff
@@ -206,6 +209,8 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
 		metadata.LatestEventsByType[parsed.Get("type").Str] = eventMetadata
 		// it's possible the latest event is a brand new room not caught by the first SELECT for joined
 		// rooms e.g when you're invited to a room so we need to make sure to set the metadata again here
+		// TODO: is this accurate now that we explicitly call NewRoomMetadata above
+		//       when handling invites?
```
kegsay, line 213: Is what accurate? The room ID? It should be if we're setting it in NewRoomMetadata?

kegsay (approved):
Good find, thanks!

state/storage.go
DMRobertson: Sorry "this" means "the comment above". I.e. _is_ it possible for us to arrive at line 202 now?

My thinking is "maybe not, but I don't have the time to investigate further and it's not broken as-is".

DMRobertson, line 213:
```suggestion
		// TODO: is the comment above now that we explicitly call NewRoomMetadata above
		//       when handling invites?
```

Issue #3, opened 2022-12-22
txn IDs: can be unset when /sync poller lags


kegsay:
If Alice and Bob are in a room, and Alice sends a message, we want to make sure that the event we pass to the client include the `transaction_id` she sent. However, Bob may get that event first down his v2 poller, meaning the database won't include that value. We fixed this in https://github.com/matrix-org/sliding-sync/commit/2920191a446f4ebbfb770ff89d67c34d0eccb1d1 to have a transaction ID cache which will inject the right transaction ID for the right user.

HOWEVER... if Alice's sync stream is lagging very badly, we may not have got the `transaction_id` by the time she does a `/sync` request. In this case, what should we do?
 - Not send the event and propagate the upstream lag?
 - Send the event but without a transaction ID?

We choose the latter, but this means that under certain circumstances we will fail to send a transaction ID with an event. This is a proxy limitation, and actual server implementations of sliding sync don't have this problem.

ara4n:
just commenting to say that this causes duplicated echo on clients (c.f. https://github.com/vector-im/element-x-ios/issues/939), so i can find it in future.

@kegsay: presumably the workaround for this is to route /send via the proxy so it can tally up the txns correctly? We can't ship EX with it in the current state.

kegsay:
Persistent duplicate events is a client-side bug, as the event ID should be used to pair up as a fallback. Flickering ephemeral duplicate events are indeed this issue.

The reason why is because:
 - if /send returns first, then we see the event ID in /sync => dedupe.
 - if /sync returns first, we dupe the message, then /send returns with the event ID => dedupe.

ara4n:
my hunch would be to go for "wait for the TID from the HS", as surely for the common case it wouldn't slow things down - it'd just slow things down for the edge case where the race has happened?

kegsay:
> my hunch would be to go for "wait for the TID from the HS", as surely for the common case it wouldn't slow things down - it'd just slow things down for the edge case where the race has happened?

It depends on the lag time of individual synchrotrons e.g on matrix.org. This can be 10s of seconds. Worst case, the delay is the total latency between the fastest (which sees the msg first without a txn ID) and slowest synchrotrons (which finally sees the txn ID).

There is a third option, which is to resend the event when we realise that the event was supposed to have a txn ID. This post-fixup would provide the best of both worlds:
 - it keeps latency as low as possible on the proxy.
[...]

PR #146, opened 2023-06-08
Ensure clients see their `transaction_id`s

DMRobertson:
Suppose
- Alice has made an incremental (i.e. with `pos`) sliding sync.
- The proxy has nothing to give to Alice right now and [is waiting for a live update](https://github.com/matrix-org/sliding-sync/blob/c95ae2aa577c5cff6f279bcf4d02badaae14f415/sync3/handler/connstate_live.go#L75).
- Alice `/send`s a message into a room using a `transaction_id`.
- Bob is also in the room and using the proxy.
- Bob's poller sees Alice's message without a transaction ID.
- **This message is handed off to Alice's /sync handler and sent to Alice _without a transaction id_.**
- Alice's poller later sees the message with the transaction id.
- (Not sure if that txn_id ever arrives at Alice's client?)

The bold bullet is bad because it means that the client doesn't realise it has just received the event it sent, meaning that the "local echo" flickers.

We want to fix this by never notifiying a waiting sync handler about Things until its dedicated poller has seen that Thing. Do this by:

1. Introduce a buffer to the API machinery which holds events which might have a transaction ID, but don't. Have entries leave the buffer (and proceed to conns) after at most a second. Stack up new message events behind it in the bugger.
2. Change the poller to track which devices have seen an event for their user which lacks a transaction ID.
3. If a poller sees a transaction ID, or if all pollers see no transaction ID, emit a payload telling the API machinery to stop waiting for a transaction ID.
4. When the API machinery receives that payload, clear up to its corresponding event in the buffer.

### DIFF
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -204,7 +204,7 @@ func (h *PollerMap) NumPollers() (count int) {
 func (h *PollerMap) deviceIDs(userID string) []string {
 	h.pollerMu.Lock()
 	defer h.pollerMu.Unlock()
-	devices := make([]string, 0)
+	var devices []string
 	for _, p := range h.Pollers {
 		if !p.terminated.Load() && p.userID == userID {
 			devices = append(devices, p.deviceID)
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -64,6 +64,7 @@ type IPollerMap interface {
 	EnsurePolling(pid PollerID, accessToken, v2since string, isStartup bool, logger zerolog.Logger)
 	NumPollers() int
 	Terminate()
+	MissingTxnID(eventID, userID, deviceID string) (bool, error)
 	SeenTxnID(eventID string) error
 }
 
@@ -210,6 +211,10 @@ func (h *PollerMap) deviceIDs(userID string) []string {
 		}
 	}
 	return devices
+}
+
+func (h *PollerMap) MissingTxnID(eventID, userID, deviceID string) (bool, error) {
+	return h.pendingTxnIDs.MissingTxnID(eventID, userID, deviceID)
 }
 
 func (h *PollerMap) SeenTxnID(eventID string) error {
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -64,6 +64,7 @@ type IPollerMap interface {
 	EnsurePolling(pid PollerID, accessToken, v2since string, isStartup bool, logger zerolog.Logger)
 	NumPollers() int
 	Terminate()
+	SeenTxnID(eventID string) error
 }
 
 // PollerMap is a map of device ID to Poller
@@ -72,6 +73,7 @@ type PollerMap struct {
 	callbacks                   V2DataReceiver
 	pollerMu                    *sync.Mutex
 	Pollers                     map[PollerID]*poller
+	pendingTxnIDs               *PendingTransactionIDs
 	executor                    chan func()
 	executorRunning             bool
 	processHistogramVec         *prometheus.HistogramVec
@@ -112,6 +114,7 @@ func NewPollerMap(v2Client Client, enablePrometheus bool) *PollerMap {
 		Pollers:  make(map[PollerID]*poller),
 		executor: make(chan func(), 0),
 	}
+	pm.pendingTxnIDs = NewPendingTransactionIDs(pm.deviceIDs)
 	if enablePrometheus {
 		pm.processHistogramVec = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 			Namespace: "sliding_sync",
@@ -193,6 +196,24 @@ func (h *PollerMap) NumPollers() (count int) {
 		}
 	}
 	return
+}
+
+// deviceIDs returns the slice of all devices currently being polled for by this user.
+// The return value is brand-new and is fully owned by the caller.
+func (h *PollerMap) deviceIDs(userID string) []string {
+	h.pollerMu.Lock()
+	defer h.pollerMu.Unlock()
+	devices := make([]string, 0)
+	for _, p := range h.Pollers {
+		if !p.terminated.Load() && p.userID == userID {
+			devices = append(devices, p.deviceID)
+		}
+	}
+	return devices
+}
+
+func (h *PollerMap) SeenTxnID(eventID string) error {
+	return h.pendingTxnIDs.SeenTxnID(eventID)
 }
 
 // EnsurePolling makes sure there is a poller for this device, making one if need be.
--- a/sync2/txnid.go
+++ b/sync2/txnid.go
@@ -35,7 +35,7 @@ type loaderFunc func(userID string) (deviceIDs []string)
 // To avoid the map growing without bound, we use a ttlcache and drop entries
 // after a short period of time.
 type PendingTransactionIDs struct {
-	// mu guards the pending field.
+	// mu guards the pending field. See MissingTxnID for rationale.
 	mu      sync.Mutex
 	pending *ttlcache.Cache
 	// loader should provide the list of device IDs
@@ -57,6 +57,24 @@ func NewPendingTransactionIDs(loader loaderFunc) *PendingTransactionIDs {
 // transaction ID for this event ID. Returns true if this is the first time we know
 // for sure that we'll never see a txn ID for this event.
 func (c *PendingTransactionIDs) MissingTxnID(eventID, userID, myDeviceID string) (bool, error) {
+	// While ttlcache is threadsafe, it does not provide a way to atomically update
+	// (get+set) a value, which means we are still open to races. For example:
+	//
+	//  - We have three pollers A, B, C.
+	//  - Poller A sees an event without txn id and calls MissingTxnID.
+	//  - `c.pending.Get()` fails, so we load up all device IDs: [A, B, C].
+	//  - Then `c.pending.Set()` with [B, C].
+	//  - Poller B sees the same event, also missing txn ID and calls MissingTxnID.
+	//  - Poller C does the same concurrently.
+	//
+	// If the Get+Set isn't atomic, then we might do e.g.
+	//  - B gets [B, C] and prepares to write [C].
+	//  - C gets [B, C] and prepares to write [B].
+	//  - Last writer wins. Either way, we never write [] and so never return true
+	//    (the all-clear signal.)
+	//
+	// This wouldn't be the end of the world (the API process has a maximum delay, and
+	// the ttlcache will expire the entry), but it would still be nice to avoid it.
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -86,7 +104,7 @@ func (c *PendingTransactionIDs) MissingTxnID(eventID, userID, myDeviceID string)
 // for this event.
 func (c *PendingTransactionIDs) SeenTxnID(eventID string) error {
 	c.mu.Lock()
-	c.mu.Unlock()
+	defer c.mu.Unlock()
 	return c.pending.Set(eventID, []string{})
 }
 
--- a/sync2/txnid.go
+++ b/sync2/txnid.go
@@ -11,13 +11,15 @@ import (
 type loaderFunc func(userID string) (deviceIDs []string)
 
 // PendingTransactionIDs is (conceptually) a map from event IDs to a list of device IDs.
-// Its keys the IDs of event we've seen which a) lack a transaction ID, and b) were sent
-// by one of the users we are polling for. The values are the list of the sender's
+// Its keys are the IDs of event we've seen which a) lack a transaction ID, and b) were
+// sent by one of the users we are polling for. The values are the list of the sender's
 // devices whose pollers are yet to see a transaction ID.
 //
 // If another poller sees the same event
+//
 //   - with a transaction ID, it emits a V2TransactionID payload with that ID and
 //     removes the event ID from this map.
+//
 //   - without a transaction ID, it removes the polling device ID from the values
 //     list. If the device ID list is now empty, the poller emits an "all clear"
 //     V2TransactionID payload.
@@ -88,6 +90,9 @@ func (c *PendingTransactionIDs) SeenTxnID(eventID string) error {
 	return c.pending.Set(eventID, []string{})
 }
 
+// removeDevice takes a device ID slice and returns a device ID slice with one
+// particular string removed. Assumes that the given slice has no duplicates.
+// Does not modify the given slice in situ.
 func removeDevice(device string, devices []string) ([]string, bool) {
 	for i, otherDevice := range devices {
 		if otherDevice == device {
--- a/sync2/txnid.go
+++ b/sync2/txnid.go
@@ -1,38 +1,98 @@
 package sync2
 
 import (
+	"fmt"
+	"sync"
 	"time"
 
 	"github.com/ReneKroon/ttlcache/v2"
 )
 
-type TransactionIDCache struct {
-	cache *ttlcache.Cache
+type loaderFunc func(userID string) (deviceIDs []string)
+
+// PendingTransactionIDs is (conceptually) a map from event IDs to a list of device IDs.
+// Its keys the IDs of event we've seen which a) lack a transaction ID, and b) were sent
+// by one of the users we are polling for. The values are the list of the sender's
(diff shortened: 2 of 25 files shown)


DMRobertson:
I assume everything is broken because I'm now delaying certain messages for up to a second.

sync3/handler/txn_id_waiter.go
```diff
@@ -0,0 +1,79 @@
...
+	if eventUpdate.EventData.Sender != t.userID && roomQueued {
+		t.publish(up)
+		return
+	}
+
+	// Don't bother queuing the event if the room isn't visible to the user.
+	if !t.subscribedOrVisible(roomID) {
+		t.publish(up)
+		return
+	}
```
DMRobertson, line 47: Also don't queue the message if it has a transaction ID

testutils/m/match.go
```diff
@@ -283,6 +283,15 @@ func MatchRoomSubscriptions(wantSubs map[string][]RoomMatcher) RespMatcher {
 	}
 }
 
+func MatchNoLists() RespMatcher {
```
DMRobertson, line 286: Seems unused

sync2/txnid.go
```diff
@@ -1,38 +1,98 @@
...
+		}
+	}
+	return changed && len(deviceIDs) == 0, nil
+}
+
+// SeenTxnID should be called to report that this device saw a transaction ID
+// for this event.
+func (c *PendingTransactionIDs) SeenTxnID(eventID string) error {
+	c.mu.Lock()
+	c.mu.Unlock()
```
kegsay, line 87: No-op due to lack of defer, plus ttlcache is thread safe iirc

sync2/txnid_test.go
```diff
@@ -2,54 +2,106 @@ package sync2
...
+	// We're not tracking any devices for Chris. A MissingTxnID call for him shouldn't
+	// cause anything to explode.
+	allClear, err = pending.MissingTxnID("event3", "chris", "C_unknown_device")
+	assertNoError(t, err)
+
+	// Delia.
+	// Delia is tracking four devices.
+	allClear, err = pending.MissingTxnID("event4", "delia", "D1")
+	assertNoError(t, err)
+	assertAllClear(t, allClear, false) // waiting on E2, E3 and E4
```
kegsay, line 70:
```suggestion
	assertAllClear(t, allClear, false) // waiting on D2, D3 and D4
```

sync2/txnid_test.go
```diff
@@ -2,54 +2,106 @@ package sync2
...
+
+	// One of Delia's devices, say D2, sees a txn ID for E4.
```
kegsay, line 72:
```suggestion
	// One of Delia's devices, say D2, sees a txn ID for event 4.
```

sync2/txnid_test.go
```diff
@@ -2,54 +2,106 @@ package sync2
...
+	assertNoError(t, err)
+
+	// Her second poller misses the transaction ID, but this shouldn't cause an all clear.
+	allClear, err = pending.MissingTxnID("event4", "delia", "E2")
+	assertNoError(t, err)
+	assertAllClear(t, allClear, false)
+}
+
+func assertAllClear(t *testing.T, got bool, want bool) {
+	if got != want {
```
kegsay, line 97:
```suggestion
	t.Helper()
	if got != want {
```

kegsay (commented):
First pass of sync2 code. Overall looking very sensible.

sync3/lists.go
```diff
@@ -222,6 +222,34 @@ func (s *InternalRequestLists) ListsByVisibleRoomIDs(muxedReqLists map[string]Re
 	return listsByRoomIDs
 }
 
+// Visible determines if a single room is currently visible in the given set of lists.
+func (s *InternalRequestLists) Visible(roomID string, muxedReqLists map[string]RequestList) bool {
```
kegsay, line 226: Honestly I would remove this and not have this functionality:
```
// Don't bother queuing the event if the room isn't visible to the user.
```

We have per room queues in txn_id_waiter, just feed them through ignoring the visibility and let conn state handle it. It feels like a smell to have txn id waiter know or care about visible rooms.

kegsay (approved):
Looks sensible. Needs more testing and removal of visiblity checks.

sync2/txnid.go
DMRobertson: Thanks.

ttlcache is threadsafe; that's not what the mutex is for. I think we need a Mutex here to ensure that the Get+Set in MissingTxnID is atomic. For example:

- We have three pollers A, B, C.
- Poller A sees an event without txn id and calls MissingTxnID.
  - `c.pending.Get()` fails, so we load up all device IDs: [A, B, C]
  - then `c.pending.Set()` with [B, C]
- Poller B sees the same event, also missing txn ID and calls MissingTxnID.
- Poller C does the same concurrently.

If the Get+Set isn't atomic, then we might do e.g.
- B gets [B, C] and prepares to write [C].
- C gets [B, C] and prepares to write [B].
- Last writer wins. Either way, we never write [] and so never return the all-clear signal. That would mean the API process hits its maximum grace period before returning the event to its sender.

Having written all that, the above wouldn't be the end of the world---particularly if this race is unlikely. But when I was thinking this through I was erring towards caution.

sync3/lists.go
DMRobertson: Was basing this on

> It needs some information from the conn ([roomSubscriptions](https://github.com/matrix-org/sliding-sync/blob/main/sync3/handler/connstate.go?rgh-link-date=2023-06-13T13%3A02%3A26Z#L31) to see if there is a room subscription for this event? That doesn't cover list windows though...)

from https://github.com/matrix-org/sliding-sync/issues/3#issuecomment-1589270205.

Happy to rip it out though; agree that it's smelly.

sync3/handler/connstate_test.go
```diff
@@ -107,7 +107,7 @@ func TestConnStateInitial(t *testing.T) {
 		}
 		return result
 	}
-	cs := NewConnState(userID, deviceID, userCache, globalCache, &NopExtensionHandler{}, &NopJoinTracker{}, nil, nil, 1000)
+	cs := NewConnState(userID, deviceID, userCache, globalCache, &NopExtensionHandler{}, &NopJoinTracker{}, nil, nil, 1000, time.Millisecond)
```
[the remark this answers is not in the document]
kegsay: I'd rather they were reliable and not flakey. This is what we do for integ tests as well fwiw - we force the pubsub to be synchronous so we know for sure that once the poller re-polls it has fully processed the previous poll. In reality this is not true as it shoves it onto a channel asyncly, but if we exercised that out of a desire to match real deployments we would have incredibly flakey tests as sometimes we would do test assertions too early, before the process has fully processed the data we sent.

In other words, pragmatism > idealism imo.

sync2/txnid.go
kegsay: Yup SGTM

sync2/poller.go
```diff
@@ -64,6 +64,8 @@ type IPollerMap interface {
 	EnsurePolling(pid PollerID, accessToken, v2since string, isStartup bool, logger zerolog.Logger)
 	NumPollers() int
 	Terminate()
+	MissingTxnID(eventID, userID, deviceID string) (bool, error)
+	SeenTxnID(eventID string) error
```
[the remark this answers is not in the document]
DMRobertson: True, but this is just the gadget that decides whether we should send an all-clear; it's the caller's responsibility to publish the payload with txn id. (Although I think even then we don't care what the txn_id is, because of the annotation logic.

sync2/txnid_test.go
DMRobertson: Thanks. (Enid was a late addition)

sync3/handler/txn_id_waiter.go
```diff
@@ -0,0 +1,96 @@
...
+	// We've decided to queue the event.
+	queue, exists := t.queues[ed.RoomID]
+	if !exists {
+		queue = make([]*caches.RoomEventUpdate, 0, 10)
+	}
+	// TODO: bound the queue size?
+	t.queues[ed.RoomID] = append(queue, eventUpdate)
+
+	// TODO: if t gets gced, will this function still run? If so, will things explode?
+	time.AfterFunc(t.maxDelay, func() { t.PublishUpToNID(ed.RoomID, ed.NID) })
```
DMRobertson, line 65: @kegsay could use your advice here. I guess the point is that this closure will still hold a reference to `t`, and so it `t` won't be GCed until this timer has run?

sync3/handler/txn_id_waiter_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+	t.Log("Queues are [1, 4] and [2, 3]")
+	t.Log("Publish up to NID 3 in room 2 should yield nids [2, 3]")
+	w.PublishUpToNID("!room2", 3)
+	assertNIDs(t, published, []int64{1, 4, 2, 3})
+	assertDelayed(t, published)
+}
+
+func assertDelayed(t *testing.T, published []publishArg) {
+	for _, p := range published {
```
kegsay, line 364:
```suggestion
	t.Helper()
	for _, p := range published {
```

sync3/handler/txn_id_waiter_test.go
```diff
@@ -0,0 +1,388 @@
...
+		if !p.delayed {
+			t.Errorf("published arg with NID %d was not delayed, but we expected it to be", p.update.(*caches.RoomEventUpdate).EventData.NID)
+		}
+	}
+}
+
+func assertNIDs(t *testing.T, published []publishArg, expectedNIDs []int64) {
+	if len(published) != len(expectedNIDs) {
```
kegsay, line 372:
```suggestion
	t.Helper()
	if len(published) != len(expectedNIDs) {
```

kegsay (approved):
I'd like to see some e2e tests to try to repro this bug (it'll flakey fail but that's better than not having any e2e tests here). This ensures that we are using the right API (e.g checking the right fields, etc.

Specifically:
 - The alice/bob use case described in a comment.
 - Multiple devices for alice, where other devices see the event without a txn id.

tests-integration/timeline_test.go
```diff
@@ -689,6 +690,120 @@ func TestTimelineTxnID(t *testing.T) {
...
 
+// Like TestTimelineTxnID, but where ...
+func TestTimelineTxnIDAfterInitialSync(t *testing.T) {
+	pqString := testutils.PrepareDBConnectionString()
+	// setup code
+	v2 := runTestV2Server(t)
+	v3 := runTestServer(t, v2, pqString, slidingsync.Opts{
+		// This needs to be greater than the request timeout, which is hardcoded to a
+		// minimum of 100ms in connStateLive.liveUpdate.
+		MaxTransactionIDDelay: 200 * time.Millisecond,
```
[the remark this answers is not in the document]
DMRobertson: Exactly this; will clarify.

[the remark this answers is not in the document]
DMRobertson: Now you mention it, this tests the "I've got a txn_id" message, but it doesn't test the all-clear message. I will see if I can whip something up quickly for the latter.

tests-integration/timeline_test.go
```diff
@@ -689,6 +690,120 @@ func TestTimelineTxnID(t *testing.T) {
 	))
 }
 
+// Like TestTimelineTxnID, but where ...
+func TestTimelineTxnIDAfterInitialSync(t *testing.T) {
```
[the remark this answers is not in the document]
DMRobertson: I'm not actually sure. The only significant difference I can spot is 

- the initial poll responses have a next_batch token. I vaugely remember that the pollers were confused if they didn't have since token from a previous iteration
- the original test didn't interrogate what Alice sees if her sync request is served before her poller sees the txn_id.

I think the latter bullet is the crucial bit. Will amend.

tests-integration/timeline_test.go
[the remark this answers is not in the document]
DMRobertson: Actually I'm going to cut a corner here, since the effect is the same: a call to `PublishUpToNID(room_id, nid)`.

tests-e2e/membership_transitions_test.go
```diff
@@ -233,17 +233,18 @@ func TestInviteRejection(t *testing.T) {
 }
```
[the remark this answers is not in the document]
DMRobertson: I went for something simpler, see baa3596.

Running on that commit:

```
$ SYNCV3_SERVER=http://localhost:8888 SYNCV3_ADDR=http://localhost:8844 go test ./tests-e2e -v -count=10 -v -run TestTransactionIDsAppearWithMultiplePollers | egrep 'PASS|FAIL'
egrep: warning: egrep is obsolescent; using grep -E
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.53s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.67s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.67s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.49s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.71s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.63s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.80s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.70s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.66s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (2.65s)
PASS
```

Running on 595e1b3bdbb28bcbb838a3dceef93d6f1d271c23:

```
$ SYNCV3_SERVER=http://localhost:8888 SYNCV3_ADDR=http://localhost:8844 go test ./tests-e2e -v -count=10 -v -run TestTransactionIDsAppearWithMultiplePollers | egrep 'PASS|FAIL'
egrep: warning: egrep is obsolescent; using grep -E
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.65s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.57s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.80s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.63s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.69s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.61s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.69s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.78s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.56s)
--- FAIL: TestTransactionIDsAppearWithMultiplePollers (1.70s)
FAIL
FAIL	github.com/matrix-org/sliding-sync/tests-e2e	16.677s
FAIL
```

The 1 second difference in timing makes me wonder if something has gone wrong with the all clear mechanism. I will investigate.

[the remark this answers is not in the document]
DMRobertson: Ah, we would not emit the V2TransactionID payloads if there were no new events.

```
$ SYNCV3_SERVER=http://localhost:8888 SYNCV3_ADDR=http://localhost:8844 go test ./tests-e2e -v -count=10 -v -run TestTransactionIDsAppearWithMultiplePollers | egrep 'PASS|FAIL'
egrep: warning: egrep is obsolescent; using grep -E
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.65s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.61s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.84s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.53s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.96s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.69s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.64s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.68s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.65s)
--- PASS: TestTransactionIDsAppearWithMultiplePollers (1.63s)
PASS
```

28b11f8 contains the fix.

PR #147, opened 2023-06-08
Add accumulator concurrency tests

kegsay:


state/accumulator.go
```diff
@@ -291,131 +292,125 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
 //     to exist in the database, and the sync stream is already linearised for us.
 //   - Else it creates a new room state snapshot if the timeline contains state events (as this now represents the current state)
 //   - It adds entries to the membership log for membership events.
-func (a *Accumulator) Accumulate(roomID string, prevBatch string, timeline []json.RawMessage) (numNew int, timelineNIDs []int64, err error) {
-	if len(timeline) == 0 {
-		return 0, nil, nil
-	}
```
DMRobertson, line 297: Is it safe to remove this? If the timeline is empty, then I think we'll try to insert 0 rows into the events table. Not sure if that will cause an error?

state/accumulator_test.go
```diff
@@ -630,6 +655,73 @@ func TestCalculateNewSnapshotDupe(t *testing.T) {
...
+	}
+	db, close := connectToDB(t)
+	defer close()
+	accumulator := NewAccumulator(db)
+	_, err := accumulator.Initialise(roomID, roomEvents)
+	if err != nil {
+		t.Fatalf("failed to Initialise accumulator: %s", err)
+	}
+
+	// spin off N goroutines to insert 1,2,3,4,5 in varying batches which can appear in the wild e.g [1,2] [1,2,3,4]
```
DMRobertson, line 675: The batches we use have the property that for any two batches, one is a prefix of the other (e.g. [1,2] is a prefix of [1,2,3,4]).

Is it possible that we could race to accumulate batches without this property, e.g. batches [1,2] and [3,5]? Does that cause problems if so?

(I think we have talked about it, but it would be good to write something down here as I keep forgetting.)

state/accumulator_test.go
```diff
@@ -630,6 +655,73 @@ func TestCalculateNewSnapshotDupe(t *testing.T) {
...
+			subset := newEvents[:(i + 1)] // i=0 => [1], i=1 => [1,2], etc
+			err := sqlutil.WithTransaction(accumulator.db, func(txn *sqlx.Tx) error {
+				numNew, _, err := accumulator.Accumulate(txn, roomID, "", subset)
+				totalNumNew += numNew
+				return err
+			})
+			if err != nil {
+				t.Errorf("goroutine failed to accumulate: %s", err)
+			}
+		}(i)
```
DMRobertson, line 699: Is there a specific race we're trying to defend against here? Or is the idea to just let the goroutines run wild and prove empirically that the test outcome is race-free?

state/accumulator.go
DMRobertson: This check was moved up into Storage.Accumulate.

v3.go
```diff
@@ -83,11 +84,13 @@ func Setup(destHomeserver, postgresURI, secret string, opts Opts) (*handler2.Han
...
 	pubSub := pubsub.NewPubSub(bufferSize)
 
+	pMap := sync2.NewPollerMap(v2Client, opts.AddPrometheusMetrics)
 	// create v2 handler
-	h2, err := handler2.NewHandler(postgresURI, sync2.NewPollerMap(v2Client, opts.AddPrometheusMetrics), storev2, store, v2Client, pubSub, pubSub, opts.AddPrometheusMetrics)
+	h2, err := handler2.NewHandler(postgresURI, pMap, storev2, store, v2Client, pubSub, pubSub, opts.AddPrometheusMetrics)
 	if err != nil {
 		panic(err)
 	}
+	pMap.SetCallbacks(h2)
```
DMRobertson, line 93: Err, what's going on here?

DMRobertson: Oh,  this was moved out of NewHandler.

DMRobertson (approved):
Looks sane, but I would appreciate it if we could write down some notes on what we're defending against.

PR #149, opened 2023-06-09
Make extension scoping work with only one field

### LINKED ISSUE #148: Extension scoping doesn't correctly handle the case where only one field is provided
https://github.com/matrix-org/matrix-spec-proposals/pull/3575/commits/30e31c8c50c1e69c7794540ba4090814087a06dc was my attempt to clarify the intended semantics of the extension scoping config.

The code as written today: https://github.com/matrix-org/sliding-sync/blob/bca1b68a08edb7293f641638543732faa462cce4/sync3/extensions/extensions.go#L99-L127

If you scope to `{"lists": ["a"]}`, the extension will only process rooms visible in `a` and :warning: ignore those in your room subscriptions. It _should_ process all rooms visible in `a` **and** all those in your room subscriptions.

Similarly if you scope to `{"rooms": ["!foo:example"]}`, the extension will only process that room and will :warning:  ignore those visible in your sliding windows. It _should_ indead process that room **and** all those in your sliding windows.

cc @bnjbvr

DMRobertson:
Fixes #148.


sync3/extensions/extensions.go
```diff
@@ -27,6 +27,10 @@ type GenericRequest interface {
 	OnlyLists() []string
 	// Returns the value of the `rooms` JSON key. nil for "not specified".
 	OnlyRooms() []string
+	// InterpretAsInitial interprets this as an initial request rather than a delta, and
+	// overwrites fields accordingly. This can be useful when fields have default
+	// values, but is a little ugly. Use sparingly.
+	InterpretAsInitial()
```
DMRobertson, line 33: I think some of this comes from the fact that we use the same struct type to represent a blob of data D and a _change_ to that blob of data.

sync3/request.go
```diff
@@ -470,6 +471,25 @@ func (r *Request) ApplyDelta(nextReq *Request) (result *Request, delta *RequestD
...
+	listKeys := make([]string, 0, len(r.Lists))
+	for listKey, _ := range r.Lists {
+		listKeys = append(listKeys, listKey)
+	}
+	return listKeys
+}
+
+// SubscribedRoomIDs builds a slice containing the room IDs this request has explicitly
+// subscribed to.
+func (r *Request) SubscribedRoomIDs() []string {
```
kegsay, line 485: These won't be verified though. Use the ones in ConnState.

sync3/extensions/receipts_test.go
```diff
@@ -16,10 +16,14 @@ func TestLiveReceiptsAggregation(t *testing.T) {
 	ext := &ReceiptsRequest{
 		Core: Core{
 			Enabled: &boolTrue,
+			Lists:   []string{"*"},
+			Rooms:   []string{"*"},
```
kegsay, line 20: Surely this is the default? Why are we needing to add this to tests?

sync3/extensions/extensions.go
```diff
@@ -202,6 +226,24 @@ func (r Request) ApplyDelta(next *Request) Request {
 	return r
 }
 
+func (r *Request) InterpretAsInitial() {
```
kegsay, line 229: I don't know what this is supposed to do.

sync3/extensions/extensions.go
```diff
@@ -77,6 +81,17 @@ func (r *Core) OnlyRooms() []string {
...
+	// An omitted/nil value for lists and rooms normally means "no change".
+	// If this extension has never been specified before, nil means "all lists/rooms".
+	if r.Lists == nil {
+		r.Lists = []string{"*"}
+	}
+	if r.Rooms == nil {
+		r.Rooms = []string{"*"}
+	}
+}
+
```
DMRobertson, line 94: I did try to make GenericRequest define a method `Default` which returned a GenericRequest, and then do `Default().ApplyDelta(incomingRequest)` instead of `incomingRequest.InterpretAsInitial`. Unfortunately I couldn't get this working so deemed this good enough.

kegsay (approved):

sync3/extensions/receipts_test.go
DMRobertson: These are indeed the defaults when you send a request into the sync v3 logic (i.e. when it gets combined with the previous requests into muxedReq). This test does not do that: it is invoking `AccountDataRequest.AppendLive` directly. The typing test below is similar.

sync3/extensions/extensions.go
DMRobertson: See the comment above the declaration of this in the interface. https://github.com/matrix-org/sliding-sync/pull/149/files#r1232127907

To illustrate: suppose I receive a request whose account data extension is configured with

```json
{
   "enabled": "true"
}
```

If this is a brand-new request, the proxy should process this in two steps:

1. Use the initial sticky parameters for `lists` and `rooms` (i.e. `["*"]`).
2. Set `enabled` to `true`

If this is a follow-up to a previous request, the proxy should interpret this as follows

1. Use the _previous_ sticky values for `lists` and `rooms`.
2. Set `enabled` to `true`.

The way I wanted to handle this was something like:

- On GenericRequest, define a Default() method that returns a new GenericRequest _of the same type_, filled with initial sticky values.
- Use this in Request.ApplyDelta to generate `curr` in the brand-new case when `isNil(curr)`. Then call `curr.ApplyDelta(next)`, just like we already do in the follow-up case.

The problem is that I couldn't figure out a way to do the first bullet. It felt like I was fighting the language. So instead, `InterpretAsInitial` has some extra logic to mutate `next` in place. The intention is that 
`Default().ApplyDelta(next)` would be equal to `next` after running `next.InterpretAsInitial()`.

DMRobertson: FTAOD I'm not very happy with InterpretAsInitial. If there is some nice way to implement Default() or similar I'm all ears.

sync3/extensions/extensions.go
```diff
@@ -100,22 +115,28 @@ func (r *Core) ApplyDelta(gnext GenericRequest) {
...
+	for _, roomInScope := range roomsToMonitor {
 		if roomInScope == roomID {
 			return true
 		}
 	}
 
-	// If the room belongs to one of the lists that this extension should process, process the update.
+	// Next determine which lists the extension is monitoring.
+	listsToMonitor := r.Lists
+	if len(listsToMonitor) > 0 && listsToMonitor[0] == "*" {
```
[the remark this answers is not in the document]
kegsay: Fair enough.

kegsay (approved):

PR #153, opened 2023-06-10
Emit txnid payloads

DMRobertson:
Completely untested.

### DIFF
--- a/pubsub/v2.go
+++ b/pubsub/v2.go
@@ -13,6 +13,7 @@ const ChanV2 = "v2ch"
 type V2Listener interface {
 	Initialise(p *V2Initialise)
 	Accumulate(p *V2Accumulate)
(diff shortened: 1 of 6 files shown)


pubsub/v2.go
```diff
@@ -40,6 +41,15 @@ type V2Accumulate struct {
 
 func (*V2Accumulate) Type() string { return "V2Accumulate" }
 
+// V2TransactionID is emitted by a poller when it sees an event with a transaction ID.
+type V2TransactionID struct {
+	EventID       string
+	TransactionID string
+	NID           int64
```
kegsay, line 48: Needs a user ID.

kegsay (approved):

pubsub/v2.go
DMRobertson: and maybe a device ID too?

PR #154, opened 2023-06-12
Poller perf: parse to-device messages first

DMRobertson:
This helps latency of to-device messages (for E2EE messaging) but doesn't help poller throughput.

### DIFF
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -429,10 +429,12 @@ func (p *poller) Poll(since string) {
 		p.initialToDeviceOnly = false
 		start = time.Now()
 		failCount = 0
+		// Do the most latency-sensitive parsing first.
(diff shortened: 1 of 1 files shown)


sync2/poller.go
```diff
@@ -429,10 +429,12 @@ func (p *poller) Poll(since string) {
 		p.initialToDeviceOnly = false
 		start = time.Now()
 		failCount = 0
+		// Do the most latency-sensitive parsing first.
+		// This only helps if the executor isn't already busy.
```
DMRobertson, line 433:
```suggestion
```

Err, I think this sentence is confused and just incorrect?

DMRobertson:
I was confused; this would have minimal impact and isn't worth doing.

PR #155, opened 2023-06-12
Make each poll a task, for easier profiling

DMRobertson:
CI seems happy, which suggests I haven't obviously nuked the poller loop. Though I don't know how robust our coverage is of poller error conditions.

### DIFF
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -434,7 +434,9 @@ func (p *poller) poll(ctx context.Context, s *pollLoopState) error {
 		return fmt.Errorf("poller terminated")
 	}
 	start := time.Now()
-	resp, statusCode, err := p.client.DoSyncV2(ctx, p.accessToken, s.since, s.firstTime, p.initialToDeviceOnly)
+	spanCtx, region := internal.StartSpan(ctx, "DoSyncV2")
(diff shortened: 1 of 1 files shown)


DMRobertson:
I ran E2E tests with pprof on and could see a Poll task: 
image

Subtasks: 
image

sync2/poller.go
```diff
@@ -390,68 +397,88 @@ func (p *poller) Poll(since string) {
...
+		// requests it might force the server to do the work all over again :(
+		waitTime := 3 * time.Second
+		p.logger.Warn().Str("duration", waitTime.String()).Int("fail-count", s.failCount).Msg("Poller: waiting before next poll")
+		timeSleep(waitTime)
+	}
+	if p.terminated.Load() {
+		return fmt.Errorf("poller terminated")
+	}
+	start := time.Now()
+	resp, statusCode, err := p.client.DoSyncV2(ctx, p.accessToken, s.since, s.firstTime, p.initialToDeviceOnly)
```
kegsay, line 437: This should also be wrapped in a task, so we have visibility.

kegsay (approved):
Please use a region (not task) on the DoSyncV2 request so if we see a 50s Poll task, and the parse tasks are 20s, we can see if the remaining 30s is waiting for the response or if something untracked is causing delays.

DMRobertson:
Got it: 
image

PR #156, opened 2023-06-13
Remove global load position

kegsay:
This enables more concurrency upstream.

### DIFF
--- a/state/event_table.go
+++ b/state/event_table.go
@@ -311,6 +311,25 @@ func (t *EventTable) LatestEventInRooms(txn *sqlx.Tx, roomIDs []string, highestN
 	return
 }
 
+func (t *EventTable) LatestEventNIDInRooms(roomIDs []string, highestNID int64) (roomToNID map[string]int64, err error) {
+	// the position (event nid) may be for a random different room, so we need to find the highest nid <= this position for this room
+	var events []Event
+	err = t.db.Select(
+		&events,
+		`SELECT event_nid, room_id FROM syncv3_events
+		WHERE event_nid IN (SELECT max(event_nid) FROM syncv3_events WHERE event_nid <= $1 AND room_id = ANY($2) GROUP BY room_id)`,
+		highestNID, pq.StringArray(roomIDs),
+	)
+	if err == sql.ErrNoRows {
+		err = nil
+	}
+	roomToNID = make(map[string]int64)
+	for _, ev := range events {
+		roomToNID[ev.RoomID] = ev.NID
+	}
+	return
+}
+
 func (t *EventTable) SelectEventsBetween(txn *sqlx.Tx, roomID string, lowerExclusive, upperInclusive int64, limit int) ([]Event, error) {
 	var events []Event
 	err := txn.Select(&events, `SELECT event_nid, event FROM syncv3_events WHERE event_nid > $1 AND event_nid <= $2 AND room_id = $3 ORDER BY event_nid ASC LIMIT $4`,
--- a/state/storage.go
+++ b/state/storage.go
@@ -31,6 +31,12 @@ type StartupSnapshot struct {
 	AllJoinedMembers map[string][]string              // room_id -> [user_id]
 }
 
+type LatestEvents struct {
+	Timeline  []json.RawMessage
+	PrevBatch string
+	LatestNID int64
+}
+
 type Storage struct {
 	Accumulator       *Accumulator
 	EventsTable       *EventTable
@@ -580,16 +586,16 @@ func (s *Storage) RoomStateAfterEventPosition(ctx context.Context, roomIDs []str
 	return
 }
 
-func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64, limit int) (map[string][]json.RawMessage, map[string]string, error) {
+func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64, limit int) (map[string]*LatestEvents, error) {
 	roomIDToRanges, err := s.visibleEventNIDsBetweenForRooms(userID, roomIDs, 0, to)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-	result := make(map[string][]json.RawMessage, len(roomIDs))
-	prevBatches := make(map[string]string, len(roomIDs))
+	result := make(map[string]*LatestEvents, len(roomIDs))
 	err = sqlutil.WithTransaction(s.Accumulator.db, func(txn *sqlx.Tx) error {
 		for roomID, ranges := range roomIDToRanges {
 			var earliestEventNID int64
+			var latestEventNID int64
 			var roomEvents []json.RawMessage
 			// start at the most recent range as we want to return the most recent `limit` events
 			for i := len(ranges) - 1; i >= 0; i-- {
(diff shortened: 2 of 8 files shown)


state/event_table.go
```diff
@@ -317,6 +317,25 @@ func (t *EventTable) LatestEventInRooms(txn *sqlx.Tx, roomIDs []string, highestN
 	return
 }
 
+func (t *EventTable) LatestEventNIDInRooms(roomIDs []string, highestNID int64) (roomToNID map[string]int64, err error) {
+	// the position (event nid) may be for a random different room, so we need to find the highest nid <= this position for this room
+	var events []Event
+	err = t.db.Select(
+		&events,
+		`SELECT event_nid, room_id FROM syncv3_events
+		WHERE event_nid IN (SELECT max(event_nid) FROM syncv3_events WHERE event_nid <= $1 AND room_id = ANY($2) GROUP BY room_id)`,
```
DMRobertson, line 326: How does this query perform/what does the query planner spit out?

I think I did something similar for bump-event types at startup where it was slow... but I guess here we have a specific list of rooms, so it might be able to just pluck out the relevant rows from indices?

state/storage.go
```diff
@@ -637,7 +650,7 @@ func (s *Storage) visibleEventNIDsBetweenForRooms(userID string, roomIDs []strin
 			return nil, fmt.Errorf("VisibleEventNIDsBetweenForRooms.SelectEventsWithTypeStateKeyInRooms: %s", err)
 		}
 	}
-	joinTimingsByRoomID, err := s.determineJoinedRoomsFromMemberships(membershipEvents)
+	joinTimingsAtFromByRoomID, err := s.determineJoinedRoomsFromMemberships(membershipEvents)
```
DMRobertson, line 653: oops, thank you!

sync3/handler/connstate.go
```diff
@@ -30,8 +30,14 @@ type ConnState struct {
...
 
-	// TODO: remove this as it is unreliable when you have concurrent updates
-	loadPosition int64
+	// This is some event NID which is used to anchor any requests for room data from the database
+	// to their per-room latest NIDs. It does this by selecting the latest NID for each requested room
+	// where the NID is <= this anchor value. Note that there are no ordering guarantees here: it's
+	// possible for the anchor to be higher than room X's latest NID and for this connection to have
+	// not yet seen room X's latest NID (it'll be sitting in the live buffer). This is why it's important
+	// that ConnState DOES NOT ignore events based on this value - it must ignore events based on the real
+	// load position for the room.
```
DMRobertson, line 39: where "real load position" means "the position of the latest event the client has confirmed receipt of"?

sync3/handler/connstate.go
```diff
@@ -87,10 +95,13 @@ func NewConnState(
...
 //   - post load() we read N events, processing them a 2nd time.
 func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
-	initialLoadPosition, joinedRooms, joinTimings, err := s.globalCache.LoadJoinedRooms(ctx, s.userID)
+	initialLoadPosition, joinedRooms, joinTimings, loadPositions, err := s.globalCache.LoadJoinedRooms(ctx, s.userID)
 	if err != nil {
 		return err
 	}
+	for roomID, nid := range loadPositions {
+		s.loadPositions[roomID] = nid
+	}
```
DMRobertson, line 104: Any reason for doing this over `s.loadPositions = loadPositions`?

sync3/handler/connstate.go
```diff
@@ -143,16 +154,21 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
...
 	}
-	s.loadPosition = initialLoadPosition
+	s.anchorLoadPosition = initialLoadPosition
 	return nil
 }
 
 // OnIncomingRequest is guaranteed to be called sequentially (it's protected by a mutex in conn.go)
 func (s *ConnState) OnIncomingRequest(ctx context.Context, cid sync3.ConnID, req *sync3.Request, isInitial bool) (*sync3.Response, error) {
-	if s.loadPosition == -1 {
+	if s.anchorLoadPosition <= 0 {
```
DMRobertson, line 163: Can we write down what nonpositive values mean in the comment for anchorLoadPosition? I.e. that we're using them as a sentinel to mean the conn hasn't been loaded yet

sync3/handler/connstate_live.go
```diff
@@ -117,8 +117,13 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 	roomEventUpdate, _ := up.(*caches.RoomEventUpdate)
 	// if this is a room event update we may not want to process this if the event nid is < loadPos,
 	// as that means we have already taken it into account
-	if roomEventUpdate != nil && !roomEventUpdate.EventData.AlwaysProcess && roomEventUpdate.EventData.NID < s.loadPosition {
-		return false
+	if roomEventUpdate != nil && !roomEventUpdate.EventData.AlwaysProcess {
+		// check if we should skip this update. Do we know of this room (lp > 0) and if so, is this event
+		// behind what we've processed before?
+		lp := s.loadPositions[roomEventUpdate.RoomID()]
+		if lp > 0 && roomEventUpdate.EventData.NID < lp {
```
DMRobertson, line 124: Aside: another implicit use of the sentinel value `0` or `-1` here

sync3/handler/connstate_live.go
```diff
@@ -149,9 +154,6 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 	rooms := s.buildRooms(ctx, builder.BuildSubscriptions())
 	for roomID, room := range rooms {
 		response.Rooms[roomID] = room
-		// remember what point we snapshotted this room, incase we see live events which we have
-		// already snapshotted here.
-		s.loadPositions[roomID] = s.loadPosition
```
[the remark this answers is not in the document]
DMRobertson: Ahh, presumably https://github.com/matrix-org/sliding-sync/blob/b7194da53e1df6b30299f41a5e2eebf9a606db85/sync3/handler/connstate_live.go#L171 still exists and does this

sync3/handler/connstate.go
```diff
@@ -143,16 +154,21 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
...
 		// load() needs no ctx so drop it
 		_, region := internal.StartSpan(ctx, "load")
-		s.load(ctx, req)
+		err := s.load(ctx, req)
```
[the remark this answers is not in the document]
kegsay: Yes it has always been able to fail. The net result would be no rooms in `s.lists` I guess.

sync3/handler/connstate.go
kegsay: Nope, let's assign instead.

sync3/handler/connstate.go
```diff
@@ -99,9 +100,7 @@ func (s *ConnState) load(ctx context.Context, req *sync3.Request) error {
 	if err != nil {
 		return err
 	}
-	for roomID, nid := range loadPositions {
-		s.loadPositions[roomID] = nid
-	}
+	s.loadPositions = loadPositions
```
DMRobertson: An integration test is failing. I wonder if `loadPositions` is sometimes nil even when `err` is not nil?

DMRobertson (approved):
LGTM once that test is passing

PR #159, opened 2023-06-14
Set sensible DB conn limits

kegsay:


state/storage.go
```diff
@@ -51,6 +52,9 @@ func NewStorage(postgresURI string) *Storage {
 		// TODO: if we panic(), will sentry have a chance to flush the event?
 		logger.Panic().Err(err).Str("uri", postgresURI).Msg("failed to open SQL DB")
 	}
+	db.SetMaxOpenConns(100)
+	db.SetMaxIdleConns(80)
+	db.SetConnMaxLifetime(time.Hour)
```
DMRobertson, line 57: Slightly unfortunate that we do this in two places, rather than build the `*sqlx.DB` once and pass it through.

DMRobertson (approved):

PR #163, opened 2023-06-15
Report the running version as a metric

DMRobertson:
Allows one to display (or superimpose) the version onto grafana graphs, to cross-reference with deployments.

### DIFF
--- a/cmd/syncv3/main.go
+++ b/cmd/syncv3/main.go
@@ -7,6 +7,7 @@ import (
 	syncv3 "github.com/matrix-org/sliding-sync"
 	"github.com/matrix-org/sliding-sync/internal"
 	"github.com/matrix-org/sliding-sync/sync2"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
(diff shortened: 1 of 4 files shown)


cmd/syncv3/main.go
```diff
@@ -112,6 +113,16 @@ func main() {
 				panic(err)
 			}
 		}()
+		// Cribbed from synapse
+		// https://github.com/matrix-org/synapse/blob/d93912042191d30ff1f7aa41d9f0779a609caca8/synapse/metrics/__init__.py#L413-L421
+		// and https://www.robustperception.io/exposing-the-software-version-to-prometheus/
+		versionGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace:   "sliding-sync",
+			Name:        "build-info",
```
DMRobertson, line 121: These should be underscores. If not, runtime error.

Need to enable prometheus metrics endpoint in the E2E tests.

kegsay: We have an entire file called metrics_test.go in integration which do this.

kegsay (approved):
Pending E2E tests to check it works.

cmd/syncv3/main.go
kegsay: This could be as trivial as starting the binary with the SYNCV3_PROM env var set then ensuring /metrics 200 OKs.

kegsay (changes requested):
This isn't what I was expecting. I would rather we didn't bundle this with end-to-end tests, as it just complicates things.

I was expecting more a simple `metrics` job which built, ran and curled /metrics to verify that it returns some data.

kegsay:
For example, we currently have:
```
            - name: Check /client works
              run: ./tests-e2e/client-check.sh
              env:
                  SYNCV3_DB: user=postgres dbname=syncv3 sslmode=disable password=postgres host=localhost
                  SYNCV3_SERVER: https://matrix-client.matrix.org
                  SYNCV3_SECRET: itsasecret
```

where `client-check.sh` is:
```bash
#!/bin/bash -eu
export SYNCV3_BINDADDR=0.0.0.0:8844

# Run the binary and stop it afterwards.
./syncv3 &
SYNCV3_PID=$!
trap "kill $SYNCV3_PID" EXIT

# wait for the server to be listening, we want this endpoint to 404 instead of connrefused
until [ \
  "$(curl -s -w '%{http_code}' -o /dev/null "http://localhost:8844/idonotexist")" \
  -eq 404 ]
do
  echo 'Waiting for server to start...'
  sleep 1
done

echo 'Checking client is reachable...'
curl -f -LI "http://localhost:8844/client/"
```

We could easily extend this to check /metrics.

DMRobertson:
Oh, I've only just noticed the word

> extend

in your comment.

DMRobertson:
Gonna drop this for now.

PR #164, opened 2023-06-15
Report the syncv2 poll times

DMRobertson:
to use as a proxy for "is Synapse being slow?"


DMRobertson:
Kegan points out that this is going to block waiting for a sync response; it's not an indicator of throughput.

PR #165, opened 2023-06-16
Add China-specific deploy instructions

nisbet-hubbard:
The default golang proxy is unavailable in that geography: [goproxy/goproxy.cn](https://github.com/goproxy/goproxy.cn).

This is no doubt familiar to veteran Gophers, but adding the note should make others’ life easier – particularly given it’s a country where a distributed encryption protocol like Matrix is highly in demand.

### Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,13 @@ $ echo -n "$(openssl rand -hex 32)" > .secret # this MUST remain the same throug
 ```
 
 Compiling from source and running:
-```
+```bash
 $ go build ./cmd/syncv3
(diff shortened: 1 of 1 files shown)


kegsay:
I'm reluctant to include anything at this stage which is specific to any particular user's deployment, because there are endless combinations which can viably be included which then pollute the core of the installation instructions. This includes adjustments to GOPROXY.

As a general rule, this encompasses things like:
 - reverse proxies (e.g nginx/apache/caddy/haproxy)
 - specific installation in certain OSes/environments/architectures (e.g raspi/cloud providers/plan9/solaris/windows)
 - service management (e.g k8s/ansible/systemd)
 - extraneous supporting services (e.g pgbouncer)

Aside from it adding noise to the installation instructions, it also becomes a maintenance burden for myself because these instructions end up getting out-of-date quickly, and are typically not tested. I'd rather not add that burden at all.

PR #169, opened 2023-06-19
Treat 403s on /sync as 401s

kegsay:


DMRobertson:
Would guess the 403s come from [here](https://github.com/matrix-org/synapse/blob/e2c8458bba5ab20f84c93a6c68e293b2d304cdc0/synapse/api/auth/__init__.py#L58-L80) in Synapse

DMRobertson (approved):

PR #172, opened 2023-06-19
Test varying DB max conns

kegsay:
Typically the failure mode is where some request needs 2+ txns by accident, which causes problems if that then blocks other things from making forward progress, causing deadlocks.

### DIFF
--- a/tests-integration/db_test.go
+++ b/tests-integration/db_test.go
@@ -1,0 +1,103 @@
+package syncv3
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	syncv3 "github.com/matrix-org/sliding-sync"
+	"github.com/matrix-org/sliding-sync/sync2"
+	"github.com/matrix-org/sliding-sync/sync3"
+	"github.com/matrix-org/sliding-sync/testutils"
+	"github.com/matrix-org/sliding-sync/testutils/m"
+)
+
+// Test that the proxy works fine with low max conns. Low max conns can be a problem
+// if a request A needs 2 conns to respond and that blocks forward progress on the server,
+// and the request can only obtain 1 conn.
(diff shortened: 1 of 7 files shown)


v3.go
```diff
@@ -75,6 +79,15 @@ func Setup(destHomeserver, postgresURI, secret string, opts Opts) (*handler2.Han
 	}
 	store := state.NewStorage(postgresURI)
 	storev2 := sync2.NewStore(postgresURI, secret)
+	for _, db := range []*sqlx.DB{store.DB, storev2.DB} {
+		if opts.DBMaxConns > 0 {
+			db.SetMaxOpenConns(opts.DBMaxConns)
+			db.SetMaxIdleConns(opts.DBMaxConns)
```
DMRobertson, line 85: Think you said there was some mysql driver that advised doing this. Can you drop a link here for our future selves?

tests-integration/db_test.go
```diff
@@ -0,0 +1,98 @@
...
+			m.MatchResponse(t, res, m.MatchRoomSubscriptionsStrict(map[string][]m.RoomMatcher{
+				roomID: {
+					m.MatchRoomTimelineMostRecent(1, []json.RawMessage{dripMsg}),
+				},
+			}))
+			t.Logf("user %s has received the drip", userID)
+		}(i)
+	}
+
+	wg.Wait()
```
DMRobertson: Does this have a timeout attached to it? (I.e. will this test always promptly terminate in CI?)

kegsay: No, it'll rely on `go test -timeout`

DMRobertson (approved):

DMRobertson:
(Isn't it worth committing this?)

DMRobertson:
(Oh, you merged manually rather than via github)

PR #174, opened 2023-06-19
Reduce DB conn usage when fetching device data

DMRobertson:
slightly terrifyingly this suggests that the impl was written was not atomically doing a get-and-update?


kegsay (approved):

kegsay:
Sadly, yes.

PR #177, opened 2023-06-20
Create systemd unit file & logging directory

### LINKED ISSUE #168: Implement zeroconfig for logging
Log messages currently go to stderr. This is easy to inspect in a docker setting, but has to be reredirected in other environments. Going forward, a better solution is to use [tulir/zeroconfig](https://github.com/tulir/zeroconfig) and let homeservers decide where they go and into which file.

I’d be happy to help out if you can provide some guidance as to how you’d like it be implemented.

nisbet-hubbard:
Also partially fixes #168.

### Pull Request Checklist

### DIFF
--- a/sliding-sync.service
+++ b/sliding-sync.service
@@ -1,0 +1,34 @@
+[Unit]
+Description=sliding sync
+
+[Service]
+Type=exec
+User=syncv3
+WorkingDirectory=/usr/local/sliding-sync
+StandardError=append:/var/log/syncv3 # append: requires systemd v240 or backports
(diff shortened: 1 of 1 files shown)


kegsay:
I'm reluctant to include anything at this stage which is specific to any particular user's deployment, because there are endless combinations which can viably be included which then pollute the core of the installation instructions. This includes adjustments to GOPROXY.

As a general rule, this encompasses things like:
 - reverse proxies (e.g nginx/apache/caddy/haproxy)
 - specific installation in certain OSes/environments/architectures (e.g raspi/cloud providers/plan9/solaris/windows)
 - service management (e.g k8s/ansible/systemd)
 - extraneous supporting services (e.g pgbouncer)

Aside from it adding noise to the installation instructions, it also becomes a maintenance burden for myself because these instructions end up getting out-of-date quickly, and are typically not tested. I'd rather not add that burden at all.

PR #178, opened 2023-06-20
Debug missing roomIDs/event timestamps seen at startup

DMRobertson:
Sentry: https://sentry.tools.element.io/organizations/element/issues/70026/events/7080aeab1d0b402e987a18f26c1ac329/?project=56&query=is%3Aunresolved

```
github.com/matrix-org/sliding-sync@v0.99.4-0.20230619174814-fc1fce54e4cd/internal/errors.go in Assert at line 77
github.com/matrix-org/sliding-sync@v0.99.4-0.20230619174814-fc1fce54e4cd/sync3/caches/global.go in (*GlobalCache).Startup at line 242
github.com/matrix-org/sliding-sync@v0.99.4-0.20230619174814-fc1fce54e4cd/sync3/handler/handler.go in (*SyncLiveHandler).Startup at line 105
```

This has been seen for as long as we've had sentry reporting.

It's not clear to me if this has any negative consequences, but I would still like to understand wtf is going on here.

https://github.com/matrix-org/sliding-sync/blob/f699eabeedea2b7e00ffd5dd6a918f88ec2f90d9/sync3/caches/global.go#L222-L253

AFAICS the elements of `roomIDs` shouldbe exactly the keys of `roomIDToMetadata`. (We hold the mutex on roomIDToMetadata----I suppose something could be modifying it underneath us that's not playing by the rules and aquiring the lock.)

This suggests then that there is a duff value in roomIDToMetadata somewhere. I would like to know what the room IDs are so I can rule out there being dodgy data.

### DIFF
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -71,10 +71,18 @@ func ExpiredSessionError() *HandlerError {
 // Which then produces:
 //
 //	assertion failed: list is not empty
-func Assert(msg string, expr bool) {
+//
+// An optional debugContext map can be provided. If it is present and sentry is configured,
+// it is added as context to the sentry events generated for failed assertions.
+func Assert(msg string, expr bool, debugContext ...map[string]interface{}) {
 	assert(msg, expr)
 	if !expr {
-		sentry.CaptureException(fmt.Errorf("assertion failed: %s", msg))
+		sentry.WithScope(func(scope *sentry.Scope) {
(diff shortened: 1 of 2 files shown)


kegsay:
I suspect https://github.com/matrix-org/sliding-sync/pull/204 was the cause - specifically spaces.

kegsay (approved):

PR #183, opened 2023-06-22
Track the time before processing a request

DMRobertson:
in particular load() and setupConnection()


DMRobertson:
```
# HELP sliding_sync_api_slow_requests Counter of slow (>=50s) requests, initial or otherwise.
# TYPE sliding_sync_api_slow_requests counter
sliding_sync_api_slow_requests 45

```

sanity checked this by forcing the threshold to be > 0 instead of > 50

PR #188, opened 2023-06-27
Increase timeout of init sync to 30mn for small homeservers

MatMaul:
The default client is then only used for incremental syncs with a 30s timeout, so decrease the default client timeout to 45s.

Signed-off-by: Mathieu Velten <mathieuv@matrix.org>

### Pull Request Checklist

### DIFF
--- a/v3.go
+++ b/v3.go
@@ -73,7 +73,7 @@ func Setup(destHomeserver, postgresURI, secret string, opts Opts) (*handler2.Han
 	// Setup shared DB and HTTP client
 	v2Client := &sync2.HTTPClient{
 		Client: &http.Client{
-			Timeout: 45 * time.Second,
+			Timeout: 5 * time.Minute,
 		},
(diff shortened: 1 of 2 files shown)


v3.go
```diff
@@ -73,7 +73,7 @@ func Setup(destHomeserver, postgresURI, secret string, opts Opts) (*handler2.Han
 	// Setup shared DB and HTTP client
 	v2Client := &sync2.HTTPClient{
 		Client: &http.Client{
-			Timeout: 5 * time.Minute,
+			Timeout: 45 * time.Second,
```
kegsay, line 76:
```suggestion
			Timeout: 5 * time.Minute,
```

kegsay (approved):
I'm happy to use a longer timeout on an initial sync for HSes which do not cache sync responses (dendrite, conduit) to work more reliably.

PR #190, opened 2023-06-28
Add WorkerPool and use it for OnE2EEData

kegsay:
- Allowing unlimited concurrency on OnE2EEData causes huge spikes in DB conns when device lists change.
- Using a high, bounded amount of concurrency ensure we don't breach DB conn limits.

With unit tests.

### DIFF
--- a/internal/pool_test.go
+++ b/internal/pool_test.go
@@ -1,0 +1,186 @@
+package internal
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+// Test basic functions of WorkerPool
+func TestWorkerPool(t *testing.T) {
+	wp := NewWorkerPool(2)
+	wp.Start()
+	defer wp.Stop()
+
+	// we should process this concurrently as N=2 so it should take 1s not 2s
+	var wg sync.WaitGroup
+	wg.Add(2)
+	start := time.Now()
+	wp.Queue(func() {
+		time.Sleep(time.Second)
+		wg.Done()
+	})
+	wp.Queue(func() {
+		time.Sleep(time.Second)
+		wg.Done()
+	})
+	wg.Wait()
+	took := time.Since(start)
+	if took > 2*time.Second {
+		t.Fatalf("took %v for queued work, it should have been faster than 2s", took)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


internal/pool_test.go
```diff
@@ -0,0 +1,186 @@
...
+		wg.Done()
+	})
+	wg.Wait()
+	took := time.Since(start)
+	if took > 2*time.Second {
+		t.Fatalf("took %v for queued work, it should have been faster than 2s", took)
+	}
+}
+
+func TestWorkerPoolDoesWorkPriorToStart(t *testing.T) {
```
DMRobertson, line 34: `DoesWork` or `DoesNoWork`?

sync2/handler2/handler.go
```diff
@@ -201,27 +204,33 @@ func (h *Handler) UpdateDeviceSince(ctx context.Context, userID, deviceID, since
...
+			internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(err)
+			return
+		}
+		// remember this to notify on pubsub later
+		h.deviceDataTicker.Remember(sync2.PollerID{
+			UserID:   userID,
+			DeviceID: deviceID,
+		})
 	})
+	wg.Wait()
```
DMRobertson, line 233: took me a moment to see that this was a local WaitGroup. So we queue up the work (don't greedily consume db conns) but block this poller until it's done?

DMRobertson (approved):
Looks sane to me!

internal/pool_test.go
kegsay: Does the work scheduled eventually, but not immediately.

sync2/handler2/handler.go
kegsay: Yes, we can't just fire and forget else we could get unfair balancing between pollers e.g 1 poller could consume >1 worker, breaking the relationship between # workers and # pollers. Rather keep it such that busy devices just slow down themselves rather than everyone.

devonh (approved):

internal/pool_test.go
DMRobertson: I'm just confused by the name, we explicitly fail if we do work before Start is called.

PR #194, opened 2023-07-10
Report a metric for the size of gappy state blocks

DMRobertson:
Sanity-checked that this worked by running [TestGappyState](https://github.com/matrix-org/sliding-sync/blob/7f03a3d78d4058a0f7b2580afa1fa57c24b79838/tests-e2e/gappy_state_test.go#L12) and polling metrics:

```
$ curl -s localhost:6060/metrics | grep gappy
# HELP sliding_sync_poller_gappy_state_size Number of events in a state block during a sync v2 gappy sync
# TYPE sliding_sync_poller_gappy_state_size histogram
sliding_sync_poller_gappy_state_size_bucket{le="1"} 1
sliding_sync_poller_gappy_state_size_bucket{le="10"} 1
sliding_sync_poller_gappy_state_size_bucket{le="100"} 1
sliding_sync_poller_gappy_state_size_bucket{le="1000"} 1
sliding_sync_poller_gappy_state_size_bucket{le="10000"} 1
sliding_sync_poller_gappy_state_size_bucket{le="+Inf"} 1
sliding_sync_poller_gappy_state_size_sum 1
sliding_sync_poller_gappy_state_size_count 1
```

### DIFF
--- a/sync2/poller.go
+++ b/sync2/poller.go
@@ -447,7 +447,7 @@ func (p *poller) Poll(since string) {
 	defer func() {
 		panicErr := recover()
 		if panicErr != nil {
-			logger.Error().Str("user", p.userID).Str("device", p.deviceID).Msg(string(debug.Stack()))
+			logger.Error().Str("user", p.userID).Str("device", p.deviceID).Msgf("%s. Traceback:\n%s", panicErr, debug.Stack())
 			internal.GetSentryHubFromContextOrDefault(ctx).RecoverWithContext(ctx, panicErr)
 		}
 		p.receiver.OnTerminated(ctx, p.userID, p.deviceID)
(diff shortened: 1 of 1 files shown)


sync2/poller.go
```diff
@@ -741,3 +755,10 @@ func (p *poller) trackTimelineSize(size int, limited bool) {
 	}
 	p.timelineSizeVec.WithLabelValues(label).Observe(float64(size))
 }
+
+func (p *poller) trackGappyStateSize(size int) {
```
kegsay, line 759: You never call this?

kegsay (approved):

sync2/poller.go
DMRobertson: :facepalm: committed locally but not pushed.

PR #196, opened 2023-07-10
Batch db updates in Accumulator.Accumulate

DMRobertson:
Suppose a timeline has `M` new message events and `S` new state events.
The logic in https://github.com/matrix-org/sliding-sync/blob/c2f4b53fdde17743010e1b7ffcd1ab894c02bf24/state/accumulator.go#L368-L402 will make a lot of small DB queries:

- up to `S` selects on 378
- `S` inserts on 394
- `M+S` updates on 399.

It would be good to batch these up, because we occasionally see large gappy state timelines (`S` big) and we want to minimise the cost of handling these.

Before this change, the new TestLargeGappyStateBlock takes 9.11s; afterwards it takes 6.95s. So a ~25% improvement on my machine. Not as much as I was hoping for, but the effect might be larger in a deployment where the database communication RTT is larger. The bulk of the remaining time is spent in v2Pub.Notify... though I don't understand why???

---

I had two doubts on this.

### Is it possible to even do a batched update?

The answer seems to be "yes". Postgres [extends](https://www.postgresql.org/docs/14/sql-update.html#id-1.9.3.182.10) the SQL spec to allow an `UPDATE ... FROM...` syntax, see https://www.postgresql.org/docs/14/sql-update.html. I cribbed this from https://github.com/jmoiron/sqlx/issues/796#issuecomment-1026102022.

### Is it possible to pre-advance the snapshot ID sequence so that the application can insert snapshots with those IDs in one update?

Answer: yes. I used the same technque Synapse does, which is essentially to repeatedly call `nexval`.


state/event_table_test.go
```diff
@@ -536,6 +537,38 @@ func TestChunkify(t *testing.T) {
...
+				t.Fatalf("got %d chunks, want %d", len(chunks), len(testCase.chunkSizes))
+			}
+			eventNID := int64(0)
+			for i := 0; i < len(chunks); i++ {
+				if len(chunks[i]) != testCase.chunkSizes[i] {
+					t.Errorf("chunk %d got %d elements, want %d", i, len(chunks[i]), testCase.chunkSizes[i])
+				}
+				for j, ev := range chunks[i] {
+					if ev.NID != eventNID {
+						t.Errorf("chunk %d got wrong event in position %d: got NID %d want NID %d", i, j, ev.NID, eventNID)
```
kegsay, line 563: It would be nice to reuse the code in TestChunkify.

state/event_table.go
```diff
@@ -302,6 +302,54 @@ func (t *EventTable) UpdateBeforeSnapshotID(txn *sqlx.Tx, eventNID, snapID, repl
...
+			--)
+			VALUES (
+				CAST(:before_state_snapshot_id AS bigint),
+			    CAST(:event_replaces_nid AS bigint),
+				CAST(:event_nid AS bigint)
+			)
+		) AS u(before_state_snapshot_id, event_replaces_nid, event_nid)
+		WHERE e.event_nid = u.event_nid`, chunk)
+		if rows != nil {
+			rows.Close()
```
kegsay, line 344: Use Exec if you aren't reading the rows.

state/event_table.go
```diff
@@ -302,6 +302,54 @@ func (t *EventTable) UpdateBeforeSnapshotID(txn *sqlx.Tx, eventNID, snapID, repl
...
+		    event_replaces_nid = u.event_replaces_nid
+		FROM (
+			--)
+			VALUES (
+				CAST(:before_state_snapshot_id AS bigint),
+			    CAST(:event_replaces_nid AS bigint),
+				CAST(:event_nid AS bigint)
+			)
+		) AS u(before_state_snapshot_id, event_replaces_nid, event_nid)
+		WHERE e.event_nid = u.event_nid`, chunk)
```
kegsay, line 342: I am failing to see what this SQL is doing.

We have a chunk of `beforeSnapshotUpdate` structs. We are somehow making that addressable as `u` by passing in `chunk`, but I'm failing to see any `$1`.

Once we have this addressable as `u`, we just do a basic update which I do understand.

state/event_table_test.go
```diff
@@ -946,3 +979,110 @@ func TestEventTableSelectUnknownEventIDs(t *testing.T) {
 		}
 	}
 }
+
+func TestEventTableUpdateBeforeSnapshotIDs(t *testing.T) {
```
kegsay, line 983: You're going to need a lot more tests to convince me that the SQL is fine. Specifically:
 - `UpdateBeforeSnapshotIDs` with no `beforeSnapshotUpdate`
 - `UpdateBeforeSnapshotIDs` with exactly 1 `beforeSnapshotUpdate`
 - `UpdateBeforeSnapshotIDs` with `len(beforeSnapshotUpdate)` > chunk size.
 -  `beforeSnapshotUpdate` has some known and some unknown event NIDs.
 - `beforeSnapshotUpdate` has some duplicate event NIDs and some not duplicated.

kegsay (commented):
The SQL isn't the worst I've seen, but it is using a syntax I'm unfamiliar with. Please link me to some docs explaining what this is doing on the postgres side. Overall, I'm not opposed to something like this, as it basically allows us to batch update in a single query which is nice. That being said, we don't seem to call `UpdateBeforeSnapshotIDs` yet, so I'm curious where you think this is going to fit in.

state/event_table_test.go
DMRobertson: That's very fair. I am also suspicious of this arcane incantation

state/event_table_test.go
DMRobertson: In the medium term I'd probably move everything over to the generic version so that there's only one true Chunkify.

DMRobertson:
> That being said, we don't seem to call `UpdateBeforeSnapshotIDs` yet, so I'm curious where you think this is going to fit in.

00c97b8 roughs out the logic I have in mind. Though it looks like plenty of tests are failing, so that'll need investigation.

DMRobertson:
Rebased atop the merge #200 and force-pushed, with apologies.

state/event_table.go
```diff
@@ -84,7 +84,12 @@ func (ev *Event) ensureFieldsSetOnEvent() error {
 
 type StrippedEvents []Event
 
+// NIDs partitions the nids in this set of stripped events into the nids of membership
+// events and all others. The returned lists inherit their order from the given slice of
+// StrippedEvents, and are never nil.
 func (se StrippedEvents) NIDs() (membershipNIDs, otherNIDs []int64) {
+	membershipNIDs = make([]int64, 0, len(se))
+	otherNIDs = make([]int64, 0, len(se))
```
DMRobertson, line 92: I chose to make this return nonnil slices if `len(se) == 0` so that I didn't have to do this at the storage layer.

state/event_table.go
DMRobertson: Let's drop the `CAST`s for a moment, so that the query is

```sql
		UPDATE syncv3_events AS e
		SET before_state_snapshot_id = u.before_state_snapshot_id,
		    event_replaces_nid = u.event_replaces_nid
		FROM (
			VALUES (:before_state_snapshot_id, :event_replaces_nid, :event_nid)
		) AS u(before_state_snapshot_id, event_replaces_nid, event_nid)
		WHERE e.event_nid = u.event_nid
```

Suppose we have two updates that we want to apply, corresponding to two VALUES expressions `(bss1, replace1, nid1)` and `(bss2, replace2, nid2`). My understanding is that sqlx will transform this query into

```sql
		UPDATE syncv3_events AS e
		SET before_state_snapshot_id = u.before_state_snapshot_id,
		    event_replaces_nid = u.event_replaces_nid
		FROM (
			VALUES ($1, $2, $3), ($4, $5, $6)
		) AS u(before_state_snapshot_id, event_replaces_nid, event_nid)
		WHERE e.event_nid = u.event_nid
```

and pass the query 6 parameters `bss1, replace1, nid1, bss2, replace2, nid2` in that order.

state/event_table_test.go
DMRobertson: > beforeSnapshotUpdate has some duplicate event NIDs and some not duplicated.

I think this is essentially undefined behaviour. Quoting postgres' [docs](https://www.postgresql.org/docs/current/sql-update.html#id-1.9.3.183.8):

> When a FROM clause is present, what essentially happens is that the target table is joined to the tables mentioned in the from_item list, and each output row of the join represents an update operation for the target table. When using FROM you should ensure that the join produces at most one output row for each row to be modified. In other words, a target row shouldn't join to more than one row from the other table(s). If it does, then only one of the join rows will be used to update the target row, but which one will be used is not readily predictable.

state/event_table.go
DMRobertson: https://www.psycopg.org/docs/extras.html#psycopg2.extras.execute_values seems to suggest that it does a similar thing (generate an appropriately-sized VALUES expression). I think psycopg2 does a better job of it, too.

DMRobertson:
@kegsay did we want to land this or not? I can't remember the outcome of our discussion.

PR #198, opened 2023-07-11
Fix #192: ignore unseen old events

### LINKED ISSUE #192: Database corruption when servers backfill over federation
Consider:

- Bob on his server invites Alice to a room.
- Alice joins the room first over federation. Proxy does the right thing and sets her membership to join. There is no timeline though due to not having backfilled.
- Alice's client backfills in the room which pulls in the invite event, but the SS proxy doesn't see it as it's backfill, not /sync.
- Charlie joins the same room via SS, which makes the SS proxy see 50 timeline events, which includes the invite. As the proxy has never seen this invite event before, it assumes it is newer than the join event and inserts it, corrupting state.

To fix this:
- SELECT nids for all events prior to insertion.
- Find the lowest-index which has a NID and drop all events earlier than this as they are backfilled.
- Discard the state block when this happens, as it cannot be trusted to do the right thing now that we are dropping parts of the timeline.

kegsay:


state/accumulator_test.go
```diff
@@ -417,86 +416,6 @@ func TestAccumulatorDupeEvents(t *testing.T) {
...
-//   - Duplicate events in the state snapshot.
-//
-// We can reproduce a message event having a event_replaces_nid by doing the following:
-//   - Create a room with initial state A,C
-//   - Accumulate events D, A, B(msg). This should be impossible because we already got A initially but whatever, roll with it, blame state resets or something.
-//   - This leads to A,B being processed and D ignored if you just take the newest results.
-//
-// This can then be tested by:
-//   - Query the current room snapshot. This will include B(msg) when it shouldn't.
-func TestAccumulatorMisorderedGraceful(t *testing.T) {
```
DMRobertson, line 432: Why are we getting rid of this---is it superseded by the new test?

DMRobertson (approved):

state/accumulator_test.go
kegsay: Because the way it was trying to repro the failure mode is impossible now, and we have other tests which repro corrupt state snapshots.

PR #201, opened 2023-07-12
Add SYNCV3_MAX_DB_CONN: use it in e2e tests

kegsay:
This is designed to catch a class of SQL transaction bugs where we BEGIN a transaction and then forget to use that `txn` var, and do other things on `sql.DB` which will use a different connection.

By testing with max conns = 1 this will deadlock. We also test with max conns = 2 to try to catch more pernicious failure modes. Using max conns = 1 effectively serialises access to the database, but some bugs may only be apparent when there is some limited amount of concurrency available e.g mid-processing this event, do X. With max conns = 1 we cannot test this, which is why we also test with max conns = 2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,6 +88,10 @@ jobs:
                 if-no-files-found: error
     end_to_end:
         runs-on: ubuntu-latest
+        strategy:
+          matrix:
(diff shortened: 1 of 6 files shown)


DMRobertson (commented):
Seems sane. Looks like the tests didn't start because Synapse was being silly---might need to kick them to rerun.

PR #202, opened 2023-07-13
Use the rooms table initially when querying latest nids

kegsay:
This should be significantly faster than the previous code.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -861,6 +861,7 @@ func (s *Storage) LatestEventNIDInRooms(roomIDs []string, highestNID int64) (roo
 		if len(slowRooms) == 0 {
 			return nil // no work to do
 		}
+		logger.Warn().Int("slow_rooms", len(slowRooms)).Msg("LatestEventNIDInRooms: pos value provided is far behind the database copy, performance degraded")
 
 		slowRoomToLatestNIDs, err := s.EventsTable.LatestEventNIDInRooms(txn, slowRooms, highestNID)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


state/storage.go
```diff
@@ -838,6 +838,42 @@ func (s *Storage) AllJoinedMembers(txn *sqlx.Tx, tempTableName string) (result m
 	return result, metadata, nil
 }
 
+func (s *Storage) LatestEventNIDInRooms(roomIDs []string, highestNID int64) (roomToNID map[string]int64, err error) {
+	roomToNID = make(map[string]int64)
+	err = sqlutil.WithTransaction(s.Accumulator.db, func(txn *sqlx.Tx) error {
+		// Pull out the latest nids for all the rooms. If they are < highestNID then use them, else we need to query the
+		// events table (slow) for the latest nid in this room which is < highestNID.
```
DMRobertson, line 845: To check I understand: we hit the slow path for a room iff it's had an event persisted with NID > highestNID.

(I think the two senses of `Latest` what's confusing me. There's "what's the latest thing in the DB", but what we're actually asking for here is "what was the last thing you saw no newer than `highestPoint`?)

DMRobertson (approved):

state/storage.go
kegsay: Yes, the slow path is used if the latest event (in the db) in the room is > highest NID. And yes, we want the latest event (at this point) for each room. Most of the time these are the same, but sometimes they aren't e.g when they race.

Issue #117, opened 2023-05-17
`TestTypingRespectsExtensionScope/Can_limit_by_list_in_an_incremental_sync` flake


DMRobertson:
E.g. https://github.com/matrix-org/sliding-sync/actions/runs/5002045944/jobs/8961515049?pr=116#step:7:1140

```
      typing_test.go:246: Alice creates rooms 3 and 4. Bob joins both.
      typing_test.go:251: room3=!MiDMNUBogsPopZEYYO:synapse room4=!gFSBMUDEKWueFhaZIG:synapse
      typing_test.go:253: Bob types in rooms 3 and 4
      typing_test.go:257: Alice incremental syncs until she sees Bob typing in room 4.
      typing_test.go:258: She a window containing all rooms, and a narrower window containing last-named room only.
      typing_test.go:259: She requests typing notifications in the narrow window only.
      typing_test.go:260: She should not see Bob typing in room 3 at any point.
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
      typing_test.go:261: SlidingSyncUntil: tested response but it failed with: MatchTyping: no typing extension
[...]

PR #203, opened 2023-07-13
WIP: expose room avatars in room subscriptions

DMRobertson:
very wip do not judge me

### DIFF
--- a/internal/roomname.go
+++ b/internal/roomname.go
@@ -60,6 +60,32 @@ func NewRoomMetadata(roomID string) *RoomMetadata {
 	}
 }
 
+// CopyHeroes returns a version of the current RoomMetadata whose Heroes field is
+// a brand-new copy of the original Heroes. The return value's Heroes field can be
+// safely modified by the caller, but it is NOT safe for the caller to modify any other
+// fields.
+func (m *RoomMetadata) CopyHeroes() *RoomMetadata {
+	newMetadata := *m
+
+	// XXX: We're doing this because we end up calling RemoveHero() to omit the
+	// currently-sycning user in various places. But this seems smelly. The set of
+	// heroes in the room is a global, room-scoped fact: it is a property of the room
+	// state and nothing else, and all users see the same set of heroes.
+	//
+	// I think the data model would be cleaner if we made the hero-reading functions
+	// aware of the currently syncing user, in order to ignore them without having to
+	// change the underlying data.
+	//
+	// copy the heroes or else we may modify the same slice which would be bad :(
+	newMetadata.Heroes = make([]Hero, len(m.Heroes))
+	copy(newMetadata.Heroes, m.Heroes)
+
+	// ⚠️ NB: there are other pointer fields (e.g. PredecessorRoomID *string) or
+	// and pointer-backed fields (e.g. LatestEventsByType map[string]EventMetadata)
+	// which are not deepcopied here.
+	return &newMetadata
+}
+
 // SameRoomName checks if the fields relevant for room names have changed between the two metadatas.
 // Returns true if there are no changes.
 func (m *RoomMetadata) SameRoomName(other *RoomMetadata) bool {
--- a/internal/roomname.go
+++ b/internal/roomname.go
@@ -109,8 +109,9 @@ func (m *RoomMetadata) IsSpace() bool {
 }
 
 type Hero struct {
-	ID   string
-	Name string
+	ID     string
+	Name   string
+	Avatar string
 }
 
 func CalculateRoomName(heroInfo *RoomMetadata, maxNumNamesPerRoom int) string {
--- a/internal/roomname.go
+++ b/internal/roomname.go
@@ -13,7 +13,12 @@ type EventMetadata struct {
 	Timestamp uint64
 }
 
-// RoomMetadata holds room-scoped data. It is primarily used in two places:
+// RoomMetadata holds room-scoped data.
+// TODO: This is a lie: we sometimes remove a user U from the list of heroes
+// when calculating the sync response for that user U. Grep for `RemoveHero`.
+//
+// It is primarily used in two places:
+//
 //   - in the caches.GlobalCache, to hold the latest version of data that is consistent
 //     between all users in the room; and
 //   - in the sync3.RoomConnMetadata struct, to hold the version of data last seen by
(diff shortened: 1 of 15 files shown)


internal/roomname.go
```diff
@@ -65,6 +66,12 @@ func (m *RoomMetadata) SameRoomName(other *RoomMetadata) bool {
 		sameHeroes(m.Heroes, other.Heroes))
 }
 
+// SameRoomAvatar checks if the fields relevant for room names have changed between the two metadatas.
+// Returns true if there are no changes.
+func (m *RoomMetadata) SameRoomAvatar(other *RoomMetadata) bool {
+	return m.AvatarURL == other.AvatarURL
```
kegsay, line 72: This probably needs to check the Heroes too like we do with room names?

sync3/caches/global.go
```diff
@@ -285,6 +285,10 @@ func (c *GlobalCache) OnNewEvent(
 		if ed.StateKey != nil && *ed.StateKey == "" {
 			metadata.NameEvent = ed.Content.Get("name").Str
 		}
+	case "m.room.avatar":
+		if ed.StateKey != nil && *ed.StateKey == "" {
+			metadata.AvatarURL = ed.Content.Get("avatar_url").Str
```
kegsay, line 290: `metadata.AvatarEvent` is a) more consistent with `NameEvent` and b) clarifies that this isn't the calculated avatar URL for the room.

sync3/room.go
```diff
@@ -9,7 +9,10 @@ import (
 )
 
 type Room struct {
-	Name              string            `json:"name,omitempty"`
+	Name string `json:"name,omitempty"`
+	// Avatar is set to the MXC url of the room if it has changed, or to "none" if
+	// it has been removed. An empty string represents no change to the avatar.
+	Avatar            string            `json:"avatar,omitempty"`
```
kegsay, line 15: Should be `*string` where literal `null` means deleted.

testutils/m/match.go
```diff
@@ -139,6 +151,20 @@ func MatchRoomTimelineMostRecent(n int, events []json.RawMessage) RoomMatcher {
 	}
 }
 
+func MatchRoomTimelineEndsWithID(wantID string) RoomMatcher {
```
kegsay, line 154: Please don't add new matchers which don't add new functionality, as it pollutes the number of functions you can use for assertions. To do this, you can just do (in tests-e2e/main_test.go): `MatchRoomTimelineMostRecent(1, []Event{{ID: "foo"}})`

tests-e2e/lists_test.go
```diff
@@ -1297,3 +1297,275 @@ func TestRangeOutsideTotalRooms(t *testing.T) {
...
+	chris := registerNewUser(t)
+
+	t.Log("Alice, Bob and Chris upload an avatar.")
+	aliceAvatar := alice.UploadContent(t, smallPNG, "alice.png", "image/png")
+	bobAvatar := bob.UploadContent(t, smallPNG, "bob.png", "image/png")
+	chrisAvatar := chris.UploadContent(t, smallPNG, "chris.png", "image/png")
+
+	alice.SetAvatar(t, aliceAvatar)
+	bob.SetAvatar(t, bobAvatar)
+	chris.SetAvatar(t, chrisAvatar)
```
kegsay, line 1320: Can we sanity check that all 3 point to different mxc URIs? Else this would rather break the tests / make them misleading.

tests-e2e/lists_test.go
```diff
@@ -1297,3 +1297,275 @@ func TestRangeOutsideTotalRooms(t *testing.T) {
...
+		t.Log("Alice syncs until she sees that avatar.")
+		alice.SlidingSyncUntil(t, res.Pos, sync3.Request{}, m.MatchRoomSubscription(dmChris, m.MatchRoomAvatar(dmAvatar)))
+
+		t.Log("Chris changes his global avatar.")
+		chrisAvatar2 := chris.UploadContent(t, smallPNG, "chris2.png", "image/png")
+		chris.SetAvatar(t, chrisAvatar2)
+		if chrisAvatar == chrisAvatar2 {
+			t.Fatalf("Chris's new avatar had the same MXC URL %s", chrisAvatar)
+		}
+		chris.SetAvatar(t, chrisAvatar2)
```
kegsay, line 1532: You do this twice?

tests-e2e/lists_test.go
```diff
@@ -1297,3 +1297,275 @@ func TestRangeOutsideTotalRooms(t *testing.T) {
...
+		dmAvatar := chris.UploadContent(t, smallPNG, "dm.png", "image/png")
+		chris.SetState(t, dmChris, "m.room.avatar", "", map[string]interface{}{
+			"url": dmAvatar,
+		})
+
+		t.Log("Alice syncs until she sees that avatar.")
+		alice.SlidingSyncUntil(t, res.Pos, sync3.Request{}, m.MatchRoomSubscription(dmChris, m.MatchRoomAvatar(dmAvatar)))
+
+		t.Log("Chris changes his global avatar.")
+		chrisAvatar2 := chris.UploadContent(t, smallPNG, "chris2.png", "image/png")
```
kegsay, line 1527: Maybe `globalChrisAvatar` would be a better name?

kegsay (approved):
Very comprehensive tests, I love it.

sync3/caches/global.go
kegsay: Also this key name is wrong, it should be `url` not `avatar_url`.

testutils/m/match.go
DMRobertson: I can do. Though I'm pretty sure I've been down this road before; I naively expect that `MatchRoomTimelineMostRecent` wants me to provide a fully-specced out Event struct

kegsay: In E2E tests it's partial matches, as you don't know all the fields.

tests-e2e/lists_test.go
DMRobertson: Fixed in 7ace2d9.

tests-e2e/lists_test.go
DMRobertson: Enforced in 7ace2d9.

testutils/m/match.go
DMRobertson: e9ebe8a

sync3/caches/global.go
DMRobertson: 2de017e

DMRobertson:
SITREP:
- e2e tests flaking due to https://github.com/matrix-org/sliding-sync/issues/117 which I am inclined to ragefix
- one integration test failure, new list ops (cause unclear---maybe this is room name reordering?)

DMRobertson:
Superseded by #208

PR #204, opened 2023-07-14
bugfix: ensure metadata maps are always init'd

kegsay:
If they aren't, we can panic on nil map access. Some of this code failed to `, ok :=` the map access which would return a zero-initialised metadata which has nil maps.

Also skips adding space children to spaces that have no joined members in them.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -313,6 +313,12 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
 		return fmt.Errorf("failed to select space children: %s", err)
 	}
 	for roomID, relations := range spaceRoomToRelations {
+		if _, exists := result[roomID]; !exists {
+			// this can happen when you join a space (so it populates the spaces table) then leave the space,
+			// so there are no joined members in the space so result doesn't include the room. In this case,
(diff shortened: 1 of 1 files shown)


state/storage.go
```diff
@@ -178,6 +178,14 @@ func (s *Storage) GlobalSnapshot() (ss StartupSnapshot, err error) {
 
 // Extract hero info for all rooms. Requires a prepared snapshot in order to be called.
 func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result map[string]internal.RoomMetadata) error {
+	loadMetadata := func(roomID string) internal.RoomMetadata {
+		metadata, ok := result[roomID]
+		if !ok {
+			metadata = *internal.NewRoomMetadata(roomID)
+		}
+		return metadata
```
DMRobertson, line 186: Should we write this back to `result`? I'm leaning towards "no": I suppose it doesn't make much difference since we ought to have `_, ok := ...` checks elsewhere, and it might give the false impression that we know about this room.

DMRobertson (approved):

state/storage.go
kegsay: No, leave it to the caller.

Issue #145, opened 2023-06-08
Unexplained error `failed to load membership events: empty slice passed to 'in' query`


DMRobertson:
Internal links: https://sentry.tools.element.io/organizations/element/issues/72685/?project=56&query=is%3Aunresolved and https://sentry.tools.element.io/organizations/element/issues/71752/?project=56&query=is%3Aunresolved

```
*errors.errorString: failed to load membership events: empty slice passed to 'in' query
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/caches/user.go", line 270, in (*UserCache).LazyLoadTimelines
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/connstate.go", line 442, in (*ConnState).getInitialRoomData
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/connstate.go", line 423, in (*ConnState).buildRooms
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/connstate.go", line 157, in (*ConnState).onIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/connstate.go", line 126, in (*ConnState).OnIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/conn.go", line 106, in (*Conn).tryRequest
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/conn.go", line 201, in (*Conn).OnIncomingRequest
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/handler.go", line 253, in (*SyncLiveHandler).serve
  File "github.com/matrix-org/sliding-sync@v0.99.3-0.20230519135138-8c1ef09d65c0/sync3/handler/handler.go", line 166, in (*SyncLiveHandler).ServeHTTP
  File "net/http/server.go", line 2109, in HandlerFunc.ServeHTTP
[...]

PR #206, opened 2023-07-17
Don't try to get data for empty room lists

S7evinK:
This should fix https://github.com/matrix-org/sliding-sync/issues/145 by not trying to fetch data for an empty room list.


kegsay (approved):
Thanks!

PR #208, opened 2023-07-18
Give clients avatar URLs for rooms

### LINKED ISSUE #166: DMs should display the opposing member's avatar
Currently in EX, avatars are only displayed for group chats. There's a proposed fix in Sliding Sync to also display them for DMs.

- add a new field to the room response which includes the calculated avatar, much like how we have `room.name`

DMRobertson:
This corresponds to the MSC change https://github.com/matrix-org/matrix-spec-proposals/pull/3575/commits/1d38101b626249ca732db54f4179b701de883a1f..912621b4d8c5bf4bc7f695acb590c055795c497f.

Supersedes #203.
Closes #166.

Includes opinionated changes to the test helpers.

Sticky bits:

- the machinery of which avatar to use and when
- having a type that json-serialises to an optional nullable string is awkward
- avoid the copy-slices-with-same-backing array problem
- a test case failure I can fix but cannot understand

I have chosen to deliberately ignore changes to the DM flag. In fact, I don't think CalculateAvatar even _reads_ the DM flag. I expect this will be good enough for a first pass.

**Strongly recommend reviewing commit-by-commit.**

### DIFF
--- a/internal/roomname.go
+++ b/internal/roomname.go
@@ -30,6 +30,7 @@ type RoomMetadata struct {
 	RoomID         string
 	Heroes         []Hero
 	NameEvent      string // the content of m.room.name, NOT the calculated name
+	AvatarEvent    string // the content of m.room.avatar, NOT the resolved avatar
 	CanonicalAlias string
 	JoinCount      int
 	InviteCount    int
--- a/internal/roomname.go
+++ b/internal/roomname.go
@@ -13,7 +13,12 @@ type EventMetadata struct {
 	Timestamp uint64
 }
 
-// RoomMetadata holds room-scoped data. It is primarily used in two places:
+// RoomMetadata holds room-scoped data.
+// TODO: This is a lie: we sometimes remove a user U from the list of heroes
+// when calculating the sync response for that user U. Grep for `RemoveHero`.
+//
+// It is primarily used in two places:
+//
 //   - in the caches.GlobalCache, to hold the latest version of data that is consistent
 //     between all users in the room; and
 //   - in the sync3.RoomConnMetadata struct, to hold the version of data last seen by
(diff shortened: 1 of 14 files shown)


internal/roomname.go
```diff
@@ -54,6 +60,32 @@ func NewRoomMetadata(roomID string) *RoomMetadata {
...
+	newMetadata := *m
+
+	// XXX: We're doing this because we end up calling RemoveHero() to omit the
+	// currently-sycning user in various places. But this seems smelly. The set of
+	// heroes in the room is a global, room-scoped fact: it is a property of the room
+	// state and nothing else, and all users see the same set of heroes.
+	//
+	// I think the data model would be cleaner if we made the hero-reading functions
+	// aware of the currently syncing user, in order to ignore them without having to
+	// change the underlying data.
```
kegsay, line 77: Then let's just do that? E.g Have a `func (m *RoomMetadata) OtherHereos(selfUserID string) []Hero` and use that throughout?

internal/roomname.go
```diff
@@ -190,3 +244,18 @@ func disambiguate(heroes []Hero) []string {
 	}
 	return disambiguatedNames
 }
+
+const noAvatar = ""
+
+// CalculateAvatar computes the avatar for the room, based on the global room metadata.
+// Assumption: metadata.RemoveHero has been called to remove the user who is syncing
+// from the list of heroes.
+func CalculateAvatar(metadata *RoomMetadata) string {
```
kegsay, line 253: This means a literal `""` is going to be treated as the same as "no avatar set". Which is probably fine.

sync3/caches/user.go
```diff
@@ -108,12 +117,15 @@ func NewInviteData(ctx context.Context, userID, roomID string, inviteState []jso
...
-					Name: j.Get("content.displayname").Str,
+					ID:     target,
+					Name:   j.Get("content.displayname").Str,
+					Avatar: j.Get("content.avatar_url").Str,
 				})
 			}
 		case "m.room.name":
 			id.NameEvent = j.Get("content.name").Str
+		case "m.room.avatar":
+			id.AvatarEvent = j.Get("content.avatar_url").Str
```
kegsay, line 128: You mean `content.url`.

testutils/m/match.go
```diff
@@ -686,13 +728,16 @@ func MatchLists(matchers map[string][]ListMatcher) RespMatcher {
...
+
 func MatchResponse(t *testing.T, res *sync3.Response, matchers ...RespMatcher) {
 	t.Helper()
 	for _, m := range matchers {
 		err := m(res)
 		if err != nil {
-			b, _ := json.Marshal(res)
-			t.Errorf("MatchResponse: %s\n%+v", err, string(b))
+			b, _ := json.MarshalIndent(res, "", "    ")
+			t.Errorf("%vMatchResponse: %s\n%s%v", AnsiRedForeground, err, string(b), AnsiResetForeground)
```
kegsay, line 740: What does this now output and why is it better?

kegsay (approved):

testutils/m/match.go
DMRobertson: It outputs the same error message but now in Red, so that I can see it easily when scrolling through hundreds of lines of test output.

It also pretty-prints the response body, which is easier to read.

testutils/m/match.go
```diff
@@ -39,6 +39,39 @@ func MatchRoomName(name string) RoomMatcher {
 	}
 }
 
+// MatchRoomAvatar builds a RoomMatcher which checks that the given room response has
+// set the room's avatar to the given value.
+func MatchRoomAvatar(wantAvatar string) RoomMatcher {
```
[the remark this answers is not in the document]
DMRobertson: I was trying to be very explicit about the different cases here, because the in-memory values and the serialised values are different, which is confusing.

sync3/caches/user.go
DMRobertson: I do, thanks. Fixed in 885b42d

internal/roomname.go
```diff
@@ -13,7 +13,12 @@ type EventMetadata struct {
 	Timestamp uint64
 }
 
-// RoomMetadata holds room-scoped data. It is primarily used in two places:
+// RoomMetadata holds room-scoped data.
+// TODO: This is a lie: we sometimes remove a user U from the list of heroes
+// when calculating the sync response for that user U. Grep for `RemoveHero`.
```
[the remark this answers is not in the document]
DMRobertson: By "room-scoped" I mean "this data is purely a function of the room". Once you have tailored it to a specific user/conn that is no-longer the case, even if the data still happens to be _about_ a room.

The net result is that when I see the type `*internal.RoomMetadata` I have no idea if it is pre- or post- tailoring.

internal/roomname.go
DMRobertson: Would love to---but was trying to keep the changes minimal and not spend too much time on this.

testutils/m/match.go
DMRobertson: Par exemple: 
image

kegsay (approved):

PR #209, opened 2023-07-18
Persist since token periodically

S7evinK:
Q: should we make the duration configurable?

As per the comments in the code, we store the first since token immediately, after that, either after the last time we stored a token is > 1min OR there are ToDevice messages in the sync response.

### DIFF
--- a/sync2/poller_test.go
+++ b/sync2/poller_test.go
@@ -277,6 +277,9 @@ func TestPollerPollFromExisting(t *testing.T) {
 			json.RawMessage(`{"event":10}`),
 		},
 	}
+	toDeviceResponses := [][]json.RawMessage{
+		{}, {}, {}, {json.RawMessage(`{}`)},
+	}
 	hasPolledSuccessfully := make(chan struct{})
 	accumulator, client := newMocks(func(authHeader, since string) (*SyncResponse, int, error) {
 		if since == "" {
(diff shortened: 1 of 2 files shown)


kegsay (commented):
LGTM, please add a test which mocks up the receiver and asserts that `UpdateDeviceSince` is:
 - called immediately the first time with the correct `since` value.
 - is not called for random events.
 - is called for to-device events.
 - is called after 1 minute (you'll need to mock this e.g `var timeSince = func(t time.Time) time.Duration` and then the test replaces `timeSince` with something that returns a high duration.

sync2/poller_test.go
```diff
@@ -343,6 +343,101 @@ func TestPollerPollFromExisting(t *testing.T) {
...
+// Check that the since token in the database
+// 1. is updated if it is the first iteration of poll
+// 2. is NOT updated for random events
+// 3. is updated if the syncV2 response contains ToDevice messages
+// 4. is updated if at least 1min has passed since we last stored a token
+func TestPollerPollUpdateDeviceSincePeriodically(t *testing.T) {
+	pid := PollerID{UserID: "@alice:localhost", DeviceID: "FOOBAR"}
+
+	syncResponses := make(chan *SyncResponse, 1)
+	accumulator, client := newMocks(func(authHeader, since string) (*SyncResponse, int, error) {
```
kegsay, line 355: Please check the `client.DoSyncV2` is called with the correct since tokens.

kegsay (approved):

kegsay (approved):

PR #213, opened 2023-07-19
Speed up startup even more

kegsay:
I realised that we are basically pulling out all joined/invited members for calculating the counts, so we might as well pull out a bit more information and use that to calculate heroes, which saves on the awful rank/partition query which took 2m45s! Byte-for-byte identical from what I can see, though I'll double check on m.org.

Also remove unused function.

With this change, startup takes 4m10s on m.org.
Without this change, startup takes 6m47s on m.org.

I have a script which JSON serialises the `StartupSnapshot` to a .json file, which I then ran through `diff`, which shows zero diffs on a static DB on my laptop. It obviously changes due to live traffic when run against m.org.


state/storage.go
```diff
@@ -938,3 +936,27 @@ func (s *Storage) Teardown() {
...
 	}
 }
+
+// circularSlice is a slice which can be appended to which will wraparound at `max`.
+// Mostly useful for lazily calculating heroes. The values returned aren't sorted.
+type circularSlice struct {
+	i    int
+	vals []int64
+	max  int
+}
```
DMRobertson, line 948: Put differently: it is a gadget into which you feed a sequence of values, outputting the `max` most recent values at the end.

state/storage_test.go
```diff
@@ -822,6 +822,56 @@ func TestAllJoinedMembers(t *testing.T) {
 	}
 }
 
+func TestCircularSlice(t *testing.T) {
+	testCases := []struct {
+		name    string
+		max     int
+		appends []int64
+		want    []int64 // these get sorted in the test
```
DMRobertson, line 830: I guess this saves you having to work out orders or mess about with modular arithmetic.

kegsay: Yep

PR #219, opened 2023-07-24
Add more metrics around connection buffers

kegsay:

PR #220, opened 2023-07-24
Protect map against concurrent map r/w

kegsay:
Comments did say to hold mu..


DMRobertson:
Introduced in #219? Or was this long-standing?

DMRobertson (approved):

PR #221, opened 2023-07-24
Adjust spam intervals

kegsay:
For now, let's see how bad it is.


DMRobertson (approved):
Fine as a stopgap for now.

PR #222, opened 2023-07-25
bugfix: fix deadlock when connections expire due to full buffers

kegsay:


sync3/connmap.go
```diff
@@ -115,9 +123,7 @@ func (m *ConnMap) Conn(cid ConnID) *Conn {
 	}
 	// e.g buffer exceeded, close it and remove it from the cache
 	logger.Info().Str("conn", cid.String()).Msg("closing connection due to dead connection (buffer full)")
-	m.mu.Lock()
```
DMRobertson, line 118: Presumably the point is that this lock would block because the mutex is already locked on -130.

DMRobertson (approved):
Is there any way we can spot these deadlocks at test time?

kegsay:
I modified a test for this particular case which acts as a regression test.

sync3/connmap.go
kegsay: Bingo

PR #224, opened 2023-07-25
Handle malformed events more gracefully

kegsay:
But handle unusual events. With regression test.

### DIFF
--- a/state/event_table.go
+++ b/state/event_table.go
@@ -57,7 +57,7 @@ func (ev *Event) ensureFieldsSetOnEvent() error {
 	}
 	if ev.Type == "" {
 		typeResult := evJSON.Get("type")
-		if !typeResult.Exists() { // empty strings for 'type' are valid apparently
+		if !typeResult.Exists() || typeResult.Type != gjson.String { // empty strings for 'type' are valid apparently
 			return fmt.Errorf("event JSON missing type key")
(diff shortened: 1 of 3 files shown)


state/event_table.go
```diff
@@ -449,14 +449,18 @@ func (c EventChunker) Subslice(i, j int) sqlutil.Chunker {
...
+func filterAndEnsureFieldsSet(events []Event) []Event {
+	result := make([]Event, 0, len(events))
 	// ensure fields are set
 	for i := range events {
 		ev := events[i]
 		if err := ev.ensureFieldsSetOnEvent(); err != nil {
-			return err
+			logger.Warn().Str("event_id", ev.ID).Err(err).Msg(
+				"filterAndEnsureFieldsSet: failed to parse event, ignoring",
+			)
```
DMRobertson, line 460: Probably want to log room ID here too.

DMRobertson (approved):
`TestEventTableMembershipDetection` is failing. Not sure if related or if it's just a flake?

You should probably also take a look at https://sentry.tools.element.io/organizations/element/issues/69559/?project=56&query=is%3Aunresolved if you haven't already.

> json: cannot unmarshal number 18446744073407678578 into Go struct field receiptInfo.content.m.read.ts of type int64

PR #225, opened 2023-07-26
Add support for DB migrations

S7evinK:
This doesn't magically resolve performance issues just yet, as we are still hitting the database quite often to do updates.
Setting a fillfactor _may_ reduce the load on the database, since the indexes possibly don't need to be updated, if there is still space in the block. (For reference - https://www.cybertec-postgresql.com/en/hot-updates-in-postgresql-for-better-performance/)

As for the claim that DeepEqual uses fewer allocations, benchmark results from my machine:

```bash
$ POSTGRES_DB=syncv3_test go test -bench=. -count 5 -run=^# -benchmem
goos: linux
goarch: amd64
pkg: github.com/matrix-org/sliding-sync/state
cpu: 12th Gen Intel(R) Core(TM) i5-12500H
BenchmarkReflectDeepEqual-16            13089988               101.9 ns/op           192 B/op          2 allocs/op
BenchmarkReflectDeepEqual-16            11241507                97.74 ns/op          192 B/op          2 allocs/op
BenchmarkReflectDeepEqual-16            12605713                97.81 ns/op          192 B/op          2 allocs/op
BenchmarkReflectDeepEqual-16            11592146               104.6 ns/op           192 B/op          2 allocs/op
BenchmarkReflectDeepEqual-16            11399110               101.0 ns/op           192 B/op          2 allocs/op
BenchmarkJSONMarhsalBytesEqual-16         866895              1161 ns/op             928 B/op         20 allocs/op
BenchmarkJSONMarhsalBytesEqual-16         975664              1180 ns/op             928 B/op         20 allocs/op
BenchmarkJSONMarhsalBytesEqual-16        1000000              1182 ns/op             928 B/op         20 allocs/op
BenchmarkJSONMarhsalBytesEqual-16        1000000              1192 ns/op             928 B/op         20 allocs/op
BenchmarkJSONMarhsalBytesEqual-16        1024558              1137 ns/op             928 B/op         20 allocs/op
PASS
```

This PR also embeds https://github.com/pressly/goose into the `syncv3` binary, allowing automatic database migrations (solving  #229)
New migrations can be created using

```bash
$ syncv3 migrate create $migrationName sql
$ syncv3 migrate create $migrationName go
```
SQL migrations work out of the box, go migrations, once the first is added, need to `import _ "github.com/matrix-org/sliding-sync/state/migrations"` in `v3.go`.

the `migrate` command has all the sub-commands mentioned on the README of goose.


state/device_data_table.go
```diff
@@ -90,7 +96,7 @@ func (t *DeviceDataTable) DeleteDevice(userID, deviceID string) error {
 }
 
 // Upsert combines what is in the database for this user|device with the partial entry `dd`
-func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (pos int64, err error) {
+func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (err error) {
```
S7evinK, line 99: Removed because unused.

state/device_data_table.go
```diff
@@ -119,11 +125,11 @@ func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (pos int64, err error)
 		if err != nil {
 			return err
 		}
-		err = txn.QueryRow(
+		_, err = txn.Exec(
 			`INSERT INTO syncv3_device_data(user_id, device_id, data) VALUES($1,$2,$3)
-			ON CONFLICT (user_id, device_id) DO UPDATE SET data=$3, id=nextval('syncv3_device_data_seq') RETURNING id`,
+			ON CONFLICT (user_id, device_id) DO UPDATE SET data=$3`,
```
S7evinK, line 130: Removed because it would conflict with HOT updates, and as above is unused.

cmd/syncv3/main.go
```diff
@@ -216,3 +229,49 @@ func WaitForShutdown(sentryInUse bool) {
...
+			fmt.Printf("\n%s must be set\n", strings.Join(requiredEnvVars, ", "))
+			os.Exit(1)
+		}
+	}
+
+	flags.Parse(os.Args[1:])
+	args := flags.Args()
+
+	if len(args) < 2 {
+		flags.Usage()
```
kegsay, line 251: Need to document usage in README or somewhere.

state/device_data_table.go
```diff
@@ -25,14 +25,15 @@ type DeviceDataTable struct {
...
 	CREATE TABLE IF NOT EXISTS syncv3_device_data (
-		id BIGINT PRIMARY KEY NOT NULL DEFAULT nextval('syncv3_device_data_seq'),
 		user_id TEXT NOT NULL,
 		device_id TEXT NOT NULL,
 		data BYTEA NOT NULL,
 		UNIQUE(user_id, device_id)
 	);
+	-- Set the fillfactor to 90%, to allow for HOT updates (e.g. we only
+	-- change the data, not anything indexed like the id)
+	ALTER TABLE syncv3_device_data SET (fillfactor = 90);
```
kegsay, line 36: Isn't this redundant now we're removing the `id` column?

kegsay (changes requested):
We need migration tests like what we do with dendrite. For now it will suffice to just use the v0.99.4 tag and then the checkout, to make sure things start up fine.

state/device_data_table.go
S7evinK: We need both, as still every `UPDATE` results in the `ctid` ("current tuple ID") to be updated by Postgres. While it is possible that there is space in the "current block" (because we "cleared" space by a different update), this raises the possibility for it.

state/device_data_table.go
```diff
@@ -67,17 +68,20 @@ func (t *DeviceDataTable) Select(userID, deviceID string, swap bool) (result *in
...
+		// json.Marshal and comparing bytes
+		if reflect.DeepEqual(writeBack, *result) {
 			// The update to the DB would be a no-op; don't bother with it.
 			// This helps reduce write usage and the contention on the unique index for
 			// the device_data table.
 			return nil
 		}
+
+		// re-marshal and write
+		data, err := json.Marshal(writeBack)
```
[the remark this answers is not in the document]
kegsay: Swapping or appending surely?

kegsay (approved):
Looks great, thank you!

state/device_data_table_test.go
```diff
@@ -193,14 +151,16 @@ func TestDeviceDataTableSwaps(t *testing.T) {
...
+		Sent: internal.ToDeviceListChangesMap([]string{"alice", "bob"}, nil),
 		New:  internal.ToDeviceListChangesMap([]string{"bob"}, []string{"charlie"}),
 	}
+	// Without swapping, we expect Alice and Bob in Sent, and Bob and Charlie in New
 	got, err = table.Select(userID, deviceID, false)
 	assertNoError(t, err)
 	assertDeviceData(t, *got, want4)
 
 	// another append then consume
+	// This results in dave to be added to Sent
```
kegsay, line 163:
```suggestion
	// This results in dave to be added to New
```

kegsay (approved):
Please land this PR and do any further changes in a different PR.

PR #226, opened 2023-07-26
Fix Nginx-example regex case-sensitivity.

olmari:
Matrix endpoints are specified to be case sensitive always, fix the Nginx-example to be case-sensitive instead case-insensitive

### Pull Request Checklist

Signed-off-by: Sami Olmari <sami@olmari.fi>


kegsay (approved):
Thanks!

PR #235, opened 2023-07-31
Do not make snapshots for lone leave events

kegsay:
Specifically this is targetting invite rejections, where the leave event is inside the leave block of the sync v2 response.

Previously, we would make a snapshot with this leave event. If the proxy wasn't in this room, it would mean the room state would just be the leave event, which is wrong. If the proxy was in the room, then state would correctly be rolled forward.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -293,7 +293,7 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
 //     to exist in the database, and the sync stream is already linearised for us.
 //   - Else it creates a new room state snapshot if the timeline contains state events (as this now represents the current state)
 //   - It adds entries to the membership log for membership events.
-func (a *Accumulator) Accumulate(txn *sqlx.Tx, roomID string, prevBatch string, timeline []json.RawMessage) (numNew int, timelineNIDs []int64, err error) {
+func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch string, timeline []json.RawMessage) (numNew int, timelineNIDs []int64, err error) {
 	// The first stage of accumulating events is mostly around validation around what the upstream HS sends us. For accumulation to work correctly
 	// we expect:
 	// - there to be no duplicate events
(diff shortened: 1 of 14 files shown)


state/accumulator.go
```diff
@@ -308,6 +308,33 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, roomID string, prevBatch string,
...
+		return 0, nil, err
+	}
+
+	// if we have just got a leave event for the polling user, and there is no snapshot for this room already, then
+	// we do NOT want to add this event to the events table, nor do we want to make a room snapshot. This is because
+	// this leave event is an invite rejection, rather than a normal event. Invite rejections cannot be processed in
+	// a normal way because we lack room state (no create event, PLs, etc). If we were to process the invite rejection,
+	// the room state would just be a single event: this leave event, which is wrong.
+	if len(dedupedEvents) == 1 && dedupedEvents[0].Type == "m.room.member" && dedupedEvents[0].Membership == "leave" &&
+		dedupedEvents[0].StateKey == userID && snapID == 0 {
```
DMRobertson, line 331: Was slightly surprised that StateKey is a string field here and not a `*string`. But I guess it's fine.

state/accumulator.go
```diff
@@ -308,6 +308,33 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, roomID string, prevBatch string,
...
+	if err != nil {
+		return 0, nil, err
+	}
+
+	// if we have just got a leave event for the polling user, and there is no snapshot for this room already, then
+	// we do NOT want to add this event to the events table, nor do we want to make a room snapshot. This is because
+	// this leave event is an invite rejection, rather than a normal event. Invite rejections cannot be processed in
+	// a normal way because we lack room state (no create event, PLs, etc). If we were to process the invite rejection,
+	// the room state would just be a single event: this leave event, which is wrong.
+	if len(dedupedEvents) == 1 && dedupedEvents[0].Type == "m.room.member" && dedupedEvents[0].Membership == "leave" &&
```
DMRobertson, line 330: Do we ever need to worry about `_leave` here?

DMRobertson (approved):
Seems sane. Is this worth an integration test? Something like

- be invited to unknown public room
- reject invite
- check nothing in the snapshots table with that room ID
- join said room
- check you get sensible state down /sync
- check there is exactly one snapshot for that room ID? (except you might get more if the timeline has state events)

state/accumulator.go
kegsay: Bizarrely, yes. The logic to determine `_` defaults `prevMembership` to `leave`, and invite rejections don't always have a `unsigned.prev_content.membership` set, so it may sometimes be unset, which defaults to leave. As the current membership will then be leave too, it will be prefixed.

state/accumulator.go
kegsay: Me too. The column itself is `state_key TEXT NOT NULL,` which sounds problematic.

sync2/poller.go
```diff
@@ -716,12 +716,23 @@ func (p *poller) parseRoomsResponse(ctx context.Context, res *SyncResponse) {
...
+		for _, ev := range roomData.Timeline.Events {
+			leaveEv := gjson.ParseBytes(ev)
+			if leaveEv.Get("content.membership").Str == "leave" && leaveEv.Get("state_key").Str == p.userID {
+				leaveEvent = ev
+				break
+			}
+		}
+		if leaveEvent != nil {
+			p.receiver.OnLeftRoom(ctx, p.userID, roomID, leaveEvent)
+		}
```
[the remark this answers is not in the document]
kegsay: I can just remove the `break` to pick the last if you'd prefer?

sync3/handler/connstate_live.go
```diff
@@ -181,11 +181,13 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
...
-			if roomEventUpdate.EventData.NID <= s.loadPositions[roomEventUpdate.RoomID()] {
-				// this update has been accounted for by the initial:true room snapshot
-				advancedPastEvent = true
+			if !roomEventUpdate.EventData.AlwaysProcess {
+				if roomEventUpdate.EventData.NID <= s.loadPositions[roomEventUpdate.RoomID()] {
+					// this update has been accounted for by the initial:true room snapshot
+					advancedPastEvent = true
+				}
+				s.loadPositions[roomEventUpdate.RoomID()] = roomEventUpdate.EventData.NID
 			}
```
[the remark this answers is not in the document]
kegsay: Without this change, an event with "always process" would _not_ be inserted in the timeline because `advancedPastEvent` would always be true due to it having a 0 nid. We do want to send down leave events in the timeline, hence wrapping `if !roomEventUpdate.EventData.AlwaysProcess {`

sync2/poller.go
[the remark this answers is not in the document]
DMRobertson: Nah, leave it as-is. I was just thinking it through for paranoia's sake.

DMRobertson (approved):

Issue #189, opened 2023-06-28
account_data returned by `/_matrix/client/unstable/org.matrix.msc3575/sync` are sometimes inconsistent


nimau:
### Description

I see a strange behaviour when updating a push rule using the push API `/_matrix/client/v3/pushrules`
Each request (`PUT` or `DELETE`) triggers a sync response but sometimes the response is incorrect.
Here, after creating a `ROOM` rule and deleting an `OVERRIDE` rule, I have 6 sync responses with different values in `account_data`.

Sometimes I have the `m.push_rules` more than once (#70, matrix-org/synapse#73), and sometimes and old value is coming back temporarily (#71 is up-to-date to date, matrix-org/synapse#75 contains the deleted rule)

Here are the requests visibles in the attached Proxyman log:

```
matrix-org/synapse#67: PUT ROOM rule (not notifying)
matrix-org/synapse#68: DELETE OVERRIDE rule
matrix-org/synapse#66: sync response: 2 rules
	ROOM + OVERRIDE
matrix-org/synapse#70: sync response: `account_data.global` contains the `m.push_rules` twice 
	ROOM + OVERRIDE
	ROOM + OVERRIDE
matrix-org/synapse#71: sync response: 1 rule
	ROOM
matrix-org/synapse#73: sync response: `account_data.global` contains the `m.push_rules` twice 
	ROOM
	ROOM
matrix-org/synapse#75: sync response: 2 rules (old data?)
	ROOM + OVERRIDE
matrix-org/synapse#77: sync response: 1 rule
	ROOM
```

[ElementX_06-28-2023-16-02-36_ss_push_rules.proxymanlogv2.zip](https://github.com/matrix-org/synapse/files/11895883/ElementX_06-28-2023-16-02-36_ss_push_rules.proxymanlogv2.zip)

(same session exported in postman format)
[ElementX_06-28-2023-16-33-17_postman.json.zip](https://github.com/matrix-org/synapse/files/11895903/ElementX_06-28-2023-16-33-17_postman.json.zip)

### Steps to reproduce

- list the steps
- that reproduce the bug
- using hyphens as bullet points

### Homeserver

matrix.org

### Synapse Version

1.86.0

### Installation Method

I don't know

### Database

I don't know

### Workers

I don't know

### Platform

I don't know

### Relevant log output

```shell
see attached files.
```

DMRobertson:
First step would be for us to reproduce this with some kind of test on the proxy side. Lines of investigation:

[...]

PR #237, opened 2023-08-02
Suppress duplicate account data events

kegsay:
With regression test.

Fixes https://github.com/matrix-org/sliding-sync/issues/189

### DIFF
--- a/sync2/handler2/handler.go
+++ b/sync2/handler2/handler.go
@@ -454,6 +454,9 @@ func (h *Handler) OnAccountData(ctx context.Context, userID, roomID string, even
 		dedupedEvents = append(dedupedEvents, events[i])
 		h.accountDataMap.Store(key, thisHash)
 	}
+	if len(dedupedEvents) == 0 {
+		return
+	}
 
 	data, err := h.Store.InsertAccountData(userID, roomID, dedupedEvents)
 	if err != nil {
--- a/sync2/handler2/handler.go
+++ b/sync2/handler2/handler.go
@@ -3,6 +3,7 @@ package handler2
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"hash/fnv"
 	"os"
 	"sync"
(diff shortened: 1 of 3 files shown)


sync2/handler2/handler.go
```diff
@@ -31,12 +32,14 @@ var logger = zerolog.New(os.Stdout).With().Timestamp().Logger().Output(zerolog.C
...
-	v2Pub     pubsub.Notifier
-	v3Sub     *pubsub.V3Sub
-	unreadMap map[string]struct {
+	pMap    sync2.IPollerMap
+	v2Store *sync2.Storage
+	Store   *state.Storage
+	v2Pub   pubsub.Notifier
+	v3Sub   *pubsub.V3Sub
+	// user_id+room_id => fnv_hash(last_event_bytes)
+	accountDataMap *sync.Map
```
DMRobertson, line 41: What is the difference between a plain map and a sync.Map? The latter has a mutex built in, and only offers concurrency-aware operations?

sync2/handler2/handler.go
```diff
@@ -433,7 +437,25 @@ func (h *Handler) UpdateUnreadCounts(ctx context.Context, roomID, userID string,
...
+			lastHash := last.(uint64)
+			if lastHash == thisHash {
+				continue // skip this event
+			}
+		}
+		dedupedEvents = append(dedupedEvents, events[i])
+		h.accountDataMap.Store(key, thisHash)
+	}
+
+	data, err := h.Store.InsertAccountData(userID, roomID, dedupedEvents)
```
DMRobertson, line 458: Suppose that E1, E2, ... are all account data events with the **same type**. This logic will behave as follows:

<table>
<th>
 <td>Input
 <td>dedupedEvents
<tr>
 <td>E1
 <td>E1
<tr>
 <td>E1, E1
 <td>E1
<tr>
 <td>E1, E2
 <td>E1, E2
<tr>
 <td>E1, E2, E1
 <td>E1, E2, E1
<tr>
 <td>E1, E1, E2, E2, E1
 <td>E1, E2, E1
</table>

I think the last three are slightly surprising since you only need to insert the most recent event with a given type. (Last write wins, per-type.)

Having written this up, I now see that there is logic for this in AccountDataTable.Insert. So maybe that's fine. (Though it's not clear to me what the responsibilities of the Handler versus the Storage versus the AccountDataTable are.)

DMRobertson (approved):
I think it's possible for a really slow poller to break this.

sync2/handler2/handler.go
kegsay: Basically yes. It is particularly good for a few use cases:
- (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or 
- (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.

In our case, we will (it's currently on the executor) do 2).

sync2/handler2/handler.go
kegsay: Yeah we already dedupe when there are dupes in the array in the storage layer, because it broke the SQL when we didn't. This is subtly different as it's across mutliple callbacks i.e it needs some in-memory structure. The handler is the obvious place for this, whereas the storage layer would be surprising.

Issue #231, opened 2023-07-27
Handle ignored user lists


kegsay:
In account data it is `m.ignored_user_list`

> To ignore a user, effectively blocking them, the client should add the target user to the m.ignored_user_list event in their account data using [/user/<user_id>/account_data/<type>](https://spec.matrix.org/v1.7/client-server-api/#put_matrixclientv3useruseridaccount_datatype). Once ignored, the client will no longer receive events sent by that user, with the exception of state events. The client should either hide previous content sent by the newly ignored user or perform a new /sync with no previous token.
>
> Invites to new rooms by ignored users will not be sent to the client. The server may optionally reject the invite on behalf of the client.

This falls under global user data, so part of the user cache. Not user room data as it isn't room specific. Connections for the ignorer need to check the sender for ignorees and if it's not a state event, or it is a state event and it's an invite from the ignoree then drop it.

Needs to be applied on `getInitialRoomData` as well as live data.

kegsay:
The caveat here is if a user ignores someone, then that causes the timelines received by the poller to be incorrect e.g instead of everyone seeing A,B,C someone sees A,C. With the logic to ignore unknown old events, all subsequent pollers would ignore B when they get A,B,C.

Unsure how often this happens in practice, but it's a proxy limitation.

DMRobertson:
- Track the ignored user list in memory and update it in response to account data changes
- Do not send messages to ignored users down timeline.
- Do not send invites from ignored users down timeline. 
- Lots of E2E tests.

DMRobertson:
Note that this only applies if Alice has ignored some events but Bob hasn't, and Bob's poller sees them first. (Otherwise Alice's poller will never see those events, because the upstream homeserver will never send them.)

PR #238, opened 2023-08-03
Suppress messages from ignored users in liveUpdate

DMRobertson:
Part one of tackling #231. The plan is described here: https://github.com/matrix-org/sliding-sync/issues/231#issuecomment-1663731782

### DIFF
--- a/sync3/handler/connstate_live.go
+++ b/sync3/handler/connstate_live.go
@@ -330,6 +330,15 @@ func (s *connStateLive) processLiveUpdateForList(
 ) (hasUpdates bool) {
 	switch update := up.(type) {
 	case *caches.RoomEventUpdate:
+		if update.EventData.StateKey == nil && s.userCache.ShouldIgnore(update.EventData.Sender) {
+			logger.Trace().
+				Str("user", s.userID).
+				Str("type", update.EventData.EventType).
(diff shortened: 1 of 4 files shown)


sync3/handler/connstate_live.go
```diff
@@ -330,6 +330,15 @@ func (s *connStateLive) processLiveUpdateForList(
 ) (hasUpdates bool) {
 	switch update := up.(type) {
 	case *caches.RoomEventUpdate:
+		if update.EventData.StateKey == nil && s.userCache.ShouldIgnore(update.EventData.Sender) {
```
kegsay, line 333: This feels like the wrong place for this check. If you have no lists for example and instead are using a direct room sub then surely we'd not do this check?

kegsay (commented):
Need to check this works with direct room subscriptions and not just lists. I don't think it will in its current form.

Otherwise LGTM, reads well, looks great :)

sync3/handler/connstate_live.go
DMRobertson: Oh true enough. Tbh I just found spotted the log line in the test output and didn't think too hard about whether this was the right place. I'll move it up to processLiveUpdate.

PR #239, opened 2023-08-03
Implement message ignoring for new conns

DMRobertson:
I've cut corners here and not tested ignored messages in lists (rather than room subscriptions) when you load up a new conn. (AFAICS they both use LazyLoadTimelines to actually get the timeline).

More of #231. Follows #238.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -37,6 +37,22 @@ type LatestEvents struct {
 	LatestNID int64
 }
 
+// DiscardIgnoredMessages modifies the struct in-place, replacing the Timeline with
+// a copy that has all ignored events omitted. The order of timelines is preserved.
+func (e *LatestEvents) DiscardIgnoredMessages(shouldIgnore func(sender string) bool) {
+	// A little bit sad to be effectively doing a copy here---most of the time there
+	// won't be any messages to ignore (and the timeline is likely short). But that copy
+	// is unlikely to be a bottleneck.
+	newTimeline := make([]json.RawMessage, 0, len(e.Timeline))
+	for _, ev := range e.Timeline {
+		parsed := gjson.ParseBytes(ev)
+		if parsed.Get("state_key").Exists() || !shouldIgnore(parsed.Get("sender").Str) {
+			newTimeline = append(newTimeline, ev)
+		}
+	}
+	e.Timeline = newTimeline
+}
+
 type Storage struct {
 	Accumulator       *Accumulator
 	EventsTable       *EventTable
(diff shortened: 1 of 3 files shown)


tests-integration/ignored_users_test.go
```diff
@@ -161,8 +161,10 @@ func TestIgnoredUsersDuringLiveUpdate(t *testing.T) {
 			},
 		},
 	})
-	t.Log("Alice sees her message.")
-	m.MatchResponse(t, aliceRes, m.MatchRoomSubscription(roomID, m.MatchRoomTimelineMostRecent(1, []json.RawMessage{aliceMsg2})))
+	t.Log("Alice sees her join, her messages and nigel's state in the timeline.")
+	m.MatchResponse(t, aliceRes, m.MatchRoomSubscription(roomID, m.MatchRoomTimeline([]json.RawMessage{
+		aliceJoin, aliceMsg, nigelState, aliceMsg2,
+	})))
```
DMRobertson, line 167: This could probably be an E2E test, to be honest. Sync to get Nigel's messages, then ignore him, then make a new conn. You shouldn't see his messages.

DMRobertson: Oh, but I guess that will only take effect on a restart, as the events will be kept around in UserRoomData timelines.

kegsay: There is no in-memory cache of timelines anymore.

state/storage.go
```diff
@@ -37,6 +37,22 @@ type LatestEvents struct {
...
+// DiscardIgnoredMessages modifies the struct in-place, replacing the Timeline with
+// a copy that has all ignored events omitted. The order of timelines is preserved.
+func (e *LatestEvents) DiscardIgnoredMessages(shouldIgnore func(sender string) bool) {
+	// A little bit sad to be effectively doing a copy here---most of the time there
+	// won't be any messages to ignore (and the timeline is likely short). But that copy
+	// is unlikely to be a bottleneck.
+	newTimeline := make([]json.RawMessage, 0, len(e.Timeline))
+	for _, ev := range e.Timeline {
+		parsed := gjson.ParseBytes(ev)
+		if parsed.Get("state_key").Exists() || !shouldIgnore(parsed.Get("sender").Str) {
```
kegsay, line 49: It would be slightly faster to check shouldIgnore first and then if true, pull out the state key, as `ParseBytes` isn't cheap.

kegsay (approved):

state/storage.go
DMRobertson: Unfortunately, we need to parse the event to get the sender.

PR #243, opened 2023-08-09
WIP: Run unit & integration tests with `go test -race`

DMRobertson:


kegsay (commented):
If CI is happy, LGTM. What a faff. It would be easier to just use an `atomic.Bool` for closing perhaps?

DMRobertson:
> What a faff.

Yes. I think I'm holding something wrong here.

According to https://go.dev/tour/concurrency/4 (I wish there was a more authoritative reference):

>  Another note: Channels aren't like files; you don't usually need to close them. Closing is only necessary when the receiver must be told there are no more values coming, such as to terminate a range loop.

I think The Right Thing might be to make PubSub.Close not close channels?

kegsay:
If you don't close channels and the channel is spread over >1 goroutine, then the goroutine will leak in cases where you have:
```go
go func() {
    for val := range ch {
    }
}()
```
as the `for` loop only terminates when the channel is closed.

kegsay:
I did half of this in #337 - need to go through the data races in integ tests.

kegsay:
https://github.com/matrix-org/sliding-sync/pull/410

PR #244, opened 2023-08-10
Fix DB migration by pulling out everything from the DB first

S7evinK:
Given the discussed issue from yesterday, this is merely a workaround for the migration, as I couldn't find a viable way to do this in SQL directly.

### DIFF
--- a/state/migrations/20230802121023_device_data_jsonb.go
+++ b/state/migrations/20230802121023_device_data_jsonb.go
@@ -39,6 +39,7 @@ func upJSONB(ctx context.Context, tx *sql.Tx) error {
 	if err != nil {
 		return err
 	}
+	defer rows.Close()
 
 	// abusing PollerID here
 	var deviceData sync2.PollerID
@@ -58,6 +59,10 @@ func upJSONB(ctx context.Context, tx *sql.Tx) error {
 		if err != nil {
 			return err
 		}
+	}
+
+	if rows.Err() != nil {
+		return rows.Err()
 	}
 
 	_, err = tx.ExecContext(ctx, "ALTER TABLE IF EXISTS syncv3_device_data DROP COLUMN IF EXISTS data;")
--- a/state/migrations/20230802121023_device_data_jsonb.go
+++ b/state/migrations/20230802121023_device_data_jsonb.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"strings"
 
+	"github.com/matrix-org/sliding-sync/sync2"
 	"github.com/pressly/goose/v3"
 )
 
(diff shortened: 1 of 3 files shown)


state/migrations/20230802121023_device_data_jsonb.go
```diff
@@ -33,10 +34,32 @@ func upJSONB(ctx context.Context, tx *sql.Tx) error {
 	if err != nil {
 		return err
 	}
-	_, err = tx.ExecContext(ctx, "UPDATE syncv3_device_data SET dataj = encode(data, 'escape')::JSONB;")
+
+	rows, err := tx.Query("SELECT user_id, device_id, data FROM syncv3_device_data")
```
kegsay, line 38: Missing defer rows close

state/migrations/20230802121023_device_data_jsonb.go
```diff
@@ -33,10 +34,32 @@ func upJSONB(ctx context.Context, tx *sql.Tx) error {
...
+		if err = rows.Scan(&deviceData.UserID, &deviceData.DeviceID, &data); err != nil {
+			return err
+		}
+		deviceDatas[deviceData] = data
+	}
+
+	for dd, d := range deviceDatas {
+		_, err = tx.ExecContext(ctx, "UPDATE syncv3_device_data SET dataj = $1 WHERE user_id = $2 AND device_id = $3;", d, dd.UserID, dd.DeviceID)
+		if err != nil {
+			return err
```
kegsay, line 59: How does this help when there are problems casting to jsonb?

state/migrations/20230802121023_device_data_jsonb_test.go
```diff
@@ -0,0 +1,86 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tx.Commit()
+
+	// insert some "invalid" data
+	dd := internal.DeviceData{
+		DeviceLists: internal.DeviceLists{
+			New:  map[string]int{"@💣:localhost": 1},
```
kegsay, line 59: It's not the unicode that's the problem on its own I think. It's the escaping part in postgres that it fails on. Can you check that this string does fail when escaped?

kegsay (changes requested):

state/migrations/20230802121023_device_data_jsonb.go
S7evinK: It's not the casting to jsonb, it's the `encode(data, 'escape')` which adds e.g. `\3` which then can't be casted to jsonb

state/migrations/20230802121023_device_data_jsonb_test.go
S7evinK: We're not using any escaping now, it is just an issue with the migration itself

PR #246, opened 2023-08-10
Optimise upsert queries for device data

kegsay:
So we don't churn the DB so much.

### DIFF
--- a/state/device_data_table.go
+++ b/state/device_data_table.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/getsentry/sentry-go"
 	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
 	"github.com/matrix-org/sliding-sync/internal"
@@ -181,8 +182,10 @@ func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (err error) {
 				return fmt.Errorf("failed to set changed bit: %w", err)
 			}
 		}
-
 		return nil
 	})
+	if err != nil && err != sql.ErrNoRows {
+		sentry.CaptureException(err)
+	}
 	return
 }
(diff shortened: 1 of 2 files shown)


state/device_data_table.go
```diff
@@ -97,55 +98,90 @@ func (t *DeviceDataTable) DeleteDevice(userID, deviceID string) error {
...
-		var row DeviceDataRow
-		err = txn.Get(&row, `SELECT data FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2`, dd.UserID, dd.DeviceID)
+		// see if anything already exists. We do this by pulling out the changed bit because we need it to calculate the new bitset,
+		// so it's a useful proxy for "does this data exist for this user/device?"
+		// We use FOR UPDATE to block any potential changes to this JSON structure through the lifetime of the transaction.
+		var changedBits int
+		err = txn.QueryRow(`SELECT data->>'c' FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2 FOR UPDATE`, dd.UserID, dd.DeviceID).Scan(&changedBits)
 		if err != nil && err != sql.ErrNoRows {
-			return err
+			return fmt.Errorf("failed to check if device data exists: %v", err)
```
S7evinK, line 107: Minor nitpick: Please use `%w` instead of `%v` so things like `errors.Is` etc. can work. (Not that we're using it here atm)
Same for the other `fmt.Errorf`

S7evinK (approved):

state/device_data_table.go
```diff
@@ -97,55 +98,90 @@ func (t *DeviceDataTable) DeleteDevice(userID, deviceID string) error {
...
 		}
-		// unmarshal and combine
-		var tempDD internal.DeviceData
-		if len(row.Data) > 0 {
-			if err = json.Unmarshal(row.Data, &tempDD); err != nil {
-				return err
+		// brand new insert, do it and return
+		if err == sql.ErrNoRows {
```
S7evinK, line 110: Speaking of `errors.Is`:
```suggestion
		if errors.Is(err, sql.ErrNoRows) {
```

Issue #245, opened 2023-08-10
Passing timestamp to clients


kegsay:
Every room has a timestamp associated with it, which is the `origin_server_ts` value of the most recent event in the timeline. However, not all events are "useful". We added `bump_event_types` to only move rooms up the list if the event `type` matched one provided by the client. E.g `bump_event_types: [ "m.room.message", "m.reaction" ]` would only look at the timestamps for those event types when determining sort order. This means the most recent event isn't always taken into account, which is _good_ because we don't want things like display name changes to bump rooms up the list.

This is encapsulated in a `map[event_type]timestamp` _per room_. This information is treated as global and is represented in `RoomMetadata` here https://github.com/matrix-org/sliding-sync/blob/main/internal/roomname.go#L42

However some of these events may be old, from before you have joined. Therefore, we need to track timestamps on a per connection basis. This is keyed by `listKey` (as `bump_event_type` is per-list) here https://github.com/matrix-org/sliding-sync/blob/main/sync3/room.go#L38 which is updated when live events come in and call `SetRoom` - https://github.com/matrix-org/sliding-sync/blob/main/sync3/lists.go#L95 and https://github.com/matrix-org/sliding-sync/blob/main/sync3/handler/connstate_live.go#L293 . This map is loaded when the connection is initially made here https://github.com/matrix-org/sliding-sync/blob/main/sync3/handler/connstate.go#L125

The point of this issue is to make ` timestamp` key in the `sync3.Room` response which is set to the correct timestamp which is what the server is using to sort based on. This needs to union multiple timestamps because 1 room can be in 2 lists each with different `bump_event_types` (e.g one bumps on messages, the other room names, so 2 different timestamps). In this case, choose the highest value.

End-to-end tests should ensure that:
 - upon joining a room, you don't see timestamps from before your join event (the join event is the lowest timestamp that should appear to avoid leaking information).
 - Events which shouldn't bump do not alter the timestamp.
 - Events which should bump do alter the timestamp.
 - Multiple lists with different `bump_event_types` choose the highest value for the timestamp.

kegsay:
This landed!

PR #247, opened 2023-08-11
Add `timestamp` to room responses

S7evinK:
Attempt for https://github.com/matrix-org/sliding-sync/issues/245

### DIFF
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -626,6 +626,7 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
 			if roomListsMeta == nil {
 				break
 			}
+			
 			evMeta := roomListsMeta.LatestEventsByType[t]
 			if evMeta.Timestamp > maxTs {
 				maxTs = evMeta.Timestamp
@@ -635,8 +636,8 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
 		// If we didn't find any events which would update the timestamp
 		// use the join event timestamp instead. Also don't leak
 		// timestamp from before we joined.
-		if maxTs == 0 {
-			if roomListsMeta != nil && maxTs < roomListsMeta.JoinTiming.Timestamp {
+		if maxTs == 0 || maxTs < roomListsMeta.JoinTiming.Timestamp {
+			if roomListsMeta != nil {
 				maxTs = roomListsMeta.JoinTiming.Timestamp
 			}
 		}
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -624,6 +624,7 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
 			JoinedCount:       metadata.JoinCount,
 			InvitedCount:      &metadata.InviteCount,
 			PrevBatch:         userRoomData.RequestedLatestEvents.PrevBatch,
+			Timestamp:         metadata.LastMessageTimestamp,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


sync3/handler/connstate.go
```diff
@@ -624,6 +624,7 @@ func (s *ConnState) getInitialRoomData(ctx context.Context, roomSub sync3.RoomSu
 			JoinedCount:       metadata.JoinCount,
 			InvitedCount:      &metadata.InviteCount,
 			PrevBatch:         userRoomData.RequestedLatestEvents.PrevBatch,
+			Timestamp:         metadata.LastMessageTimestamp,
```
kegsay, line 627: You don't know if this is what they want. `getInitialRoomData` is a central point called for multiple lists: you need to know the `bump_event_types` values to know what timestamp to pull here. Add that as a param to `getInitialRoomData`

sync3/handler/connstate_live.go
```diff
@@ -187,6 +187,14 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 		// include this update in the rooms response TODO: filters on event type?
 		userRoomData := roomUpdate.UserRoomMetadata()
 		r := response.Rooms[roomUpdate.RoomID()]
+
+		conMeta := s.lists.ReadOnlyRoom(roomUpdate.RoomID())
+		for _, ts := range conMeta.LastInterestedEventTimestamps {
```
kegsay, line 192: Needs comments. This is using the highest timestamp from each list for this room I guess?

tests-e2e/timestamp_test.go
```diff
@@ -0,0 +1,126 @@
...
+					TimelineLimit: 10,
+				},
+				BumpEventTypes: []string{"m.reaction"}, // only reactions bump the timestamp
+			},
+		},
+	})
+
+	// Bob should see a different timestamp than alice, as he just joined
+	bobTimestampBefore := resBob.Rooms[roomID].Timestamp
+	if bobTimestampBefore == timestampBeforeBobJoined {
```
kegsay, line 53: Really this should equal `res.Rooms[roomID].Timestamp` as that is going to be bob's join due to :30.

tests-e2e/timestamp_test.go
```diff
@@ -0,0 +1,126 @@
...
+		Type: "m.room.message",
+		Content: map[string]interface{}{
+			"msgtype": "m.text",
+			"body":    "Hello, world!",
+		},
+	})
+
+	resBob = bob.SlidingSyncUntilEventID(t, resBob.Pos, roomID, eventID)
+	bobTimestampAfter = resBob.Rooms[roomID].Timestamp
+	if bobTimestampBefore == bobTimestampAfter {
```
kegsay, line 88: Again, we need to know exactly what timestamp this should be - use alice to figure this out.

tests-e2e/timestamp_test.go
```diff
@@ -0,0 +1,126 @@
...
+			"name": "random name",
+		},
+	})
+
+	resBob = bob.SlidingSyncUntilEventID(t, resBob.Pos, roomID, eventID)
+	bobTimestampAfter = resBob.Rooms[roomID].Timestamp
+	if bobTimestampBefore != bobTimestampAfter {
+		t.Fatalf("expected timestamps to be the same, but they aren't: %v vs %v", bobTimestampBefore, bobTimestampAfter)
+	}
+}
```
kegsay, line 126: Needs more tests. Specifically:
- If you now do an initial sync (no pos) as Bob with the same lists, I don't think you'll get the right value. You'll get the latest timestamp (m.room.name) and not the m.reaction timestamp.
- Ensure you `time.Sleep(time.Millisecond)` after each send to ensure that timestamps will change.
- If Charlie now joins the room with the same bump_event_types as bob, he should see the timestamp of his join.

kegsay (changes requested):
Almost there, thanks.

kegsay (approved):

Issue #232, opened 2023-07-27
Checkpoint room state


kegsay:
Users can join and leave rooms. If all users who are using the proxy leave a room, then Stuff happens and someone rejoins, the poller will be resent the _entire room state from fresh_. We currently do not use this correctly, and instead treat it as a "gappy sync", meaning we take what we previously had as the room state, and "roll forward" using what we are told in the state block.

We shouldn't do that, as it means that we can't handle state resets correctly (i.e we never remove state by rolling forwards), and if we calculate the wrong state in the past (not uncommon), we cannot self-heal by using the latest authoritative answer from upstream.

To do this is reasonably easy. All we need to do is, in `Initialise`:
 - is there an `m.room.create` even in the `state` block?
    * No: gappy sync, roll forward.
    * Yes: this sync contains everything, do a complete snapshot "rolling forwards" from the empty set [].

DMRobertson:
> All we need to do is

then emit a cache invalidation payload (#294)

DMRobertson:
I think what we need Initialise to do is then:

```
	// 0. If the state block is empty, bail.
	//
	// 1. Capture the current snapshot ID, possibly zero.
	//
	// 2. Determine the kind of state block. Is it:
	//    a. A complete snapshot of the room, containing m.room.create?
	//    b. B "gappy sync" state delta, not containing m.room.create?
	//
	// 3. Does this alter the current state of the room from the proxy's POV?
	//    a. Complete snapshot: state changes iff current snapshot's contents != state block.
	//       (What if this comes from a slow poller? Ahh, it'll catch up with later events and self-heal via Accumulate.)
	//    b. Gappy sync: state changes iff some state block event is not in the current snapshot.
	//    If the state hasn't altered, bail.
	//
	// 4. Create new snapshot.
	//    a. Complete snapshot: trust the state block verbatim.
	//    b. Gappy sync: build a map from (type, state_key) to nids for the current snapshot. Overwrite its entries
	//       with the new things in `state`. (There must be similar logic already in Accumulate for this?)
	//    Store the snapshot. Mark the room's current state as being this snapshot.
	//
	// 5. If the starting snapshot ID was not zero, emit a cache invalidation payload.
```

DMRobertson:
After discussion:

```
	// 0. If the state block is empty, bail.
	//
	// 1. Capture the current snapshot ID, possibly zero.
    // 
    // 2. Fetch the current state of the room, as a map from (type, state_key) to event.
[...]

PR #248, opened 2023-08-11
WIP: create new state snapshots on join

### LINKED ISSUE #211: Persist large gappy state blocks as a single snapshot
Ideally, we would instead make the events in the `state` block part of the _state snapshot_ and then add new timeline events in `Accumulate` only. This:
 - more accurately maps onto what upstream is telling us.
 - is much faster as we aren't calculating state snapshots for each 10,000s event.

This implies a third function in the accumulator between `Initialise` and `Accumulate` which can both snapshot and roll forward the timeilne in a single txn.

_Originally posted by @kegsay in https://github.com/matrix-org/sliding-sync/issues/196#issuecomment-1639902732_

Original gappy state impl in #71.

kegsay:
Fixes #211 by implementing #232.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -176,7 +176,7 @@ func (a *Accumulator) ProcessRoomEvents(userID, roomID string, timeline, state [
 		// we know there are some events to insert which are new. We need to work out what snapshot
 		// to base these changes on.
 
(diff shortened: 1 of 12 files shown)


kegsay:
Closing as dmr has ended up doing something different here.

PR #249, opened 2023-08-14
device data: use CBOR instead of JSONB

kegsay:
Using JSONB columns adds too much DB load. Prefer a slightly faster serialisation format instead, and use the old system of handling BYTEA, which is about 2x faster.
```
BenchmarkSerialiseDeviceDataJSON-12    	    1770	    576646 ns/op	  426297 B/op	    6840 allocs/op
BenchmarkSerialiseDeviceDataCBOR-12    	    4635	    247509 ns/op	  253971 B/op	    4796 allocs/op
```
This was using a growing list of 1000 device list changes.


kegsay:
In a more realistic setting, performing 100x `Upsert` then `Select` with swap=true, for 10k user IDs reduces the time taken from 3.5-3.8s to 3.0-3.1s.

S7evinK (approved):

PR #253, opened 2023-08-15
Return M_UNKNOWN_TOKEN for unknown tokens

kegsay:


tests-integration/token_management_test.go
DMRobertson, line 1: Could have been an E2E test where you invalidate the token by logging out the device.

DMRobertson (approved):

DMRobertson (approved):

PR #255, opened 2023-08-17
Ignore more stray timelines in Accumulate

DMRobertson:
Described in https://github.com/matrix-org/sliding-sync/issues/211#issuecomment-1682728834.

This is essentially cherry picked from #248, in particular the commit
    https://github.com/matrix-org/sliding-sync/pull/248/commits/8c7046e96b4efec5930e02cb9d8daefe65d3d7e7

This should prevent creating new snapshots that don't reflect the state of the room. We'll need a followup task to clean up bad snapshots.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -335,7 +335,7 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch 
 			// All okay, continue on.
 		} else {
 			// Bail out and complain loudly.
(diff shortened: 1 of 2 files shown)


DMRobertson:
I will need to

- address test failures
- see if there are any tests we want to add. Might be able to pluck from #248.

DMRobertson:
Easier than I expected to get the tests happy.

It looks like the test changes in #248 are fixups to ensuring that we never send a nonempty state block with an empty timeline. This PR targets empty state blocks and nonempty timelines, so I think that is irrelevant here.

PR #257, opened 2023-08-17
Remove incorrect `m.server` example from readme

turt2live:
There's nothing in the spec which requires this on the client .well-known, and it's improperly formatted for federation.


DMRobertson (approved):
Thanks!

Issue #256, opened 2023-08-17
Clean up bad snapshots from existing databases.


DMRobertson:
Followup to https://github.com/matrix-org/sliding-sync/pull/255 and https://github.com/matrix-org/sliding-sync/issues/211. 

Kegan has some SQL prepared for this in a markdown document. We can use Till's migration machinery (#225, thanks Till) to run this script as a one-off.

PR #266, opened 2023-08-22
Initialise: don't snapshot without create event

DMRobertson:
Followup to #255. Prerequisite for #256.

We don't want there to be "bogus" snapshots without a create event. We used to create such snapshots if we were informed of a stray state event in a room timeline. #235 targeted on known occurrance; #255 stopped this more generally.

We noticed another way to create bogus snapshots today. Namely: if the first time the proxy saw a room it had a state block lacking a create event. We don't think Synapse has been sending us these (outside of a gappy sync, where it is expected). But let's err towards caution to try and keep the database sane.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -217,7 +217,7 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
 		// single stray event by looking for the create event.
 		hasCreate := false
 		for _, e := range events {
-			if e.Type == "m.room.create" {
+			if e.Type == "m.room.create" && e.StateKey == "" {
 				hasCreate = true
 				break
 			}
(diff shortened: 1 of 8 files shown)


kegsay (approved):
I suspect there are integration tests which do not ensure this.

DMRobertson:
> I suspect there are integration tests which do not ensure this.

d3285a3 cherry-picked from #248 which may help with this, though surprisingly only an unrelated accumulator test failed.

state/accumulator.go
```diff
@@ -212,6 +212,33 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
...
+					"room_id":   roomID,
+					"len_state": len(events),
+				})
+				sentry.CaptureMessage(errMsg)
+			})
+			logger.Warn().
+				Str("room_id", roomID).
+				Int("len_state", len(events)).
+				Msg(errMsg)
+			return fmt.Errorf(errMsg)
```
DMRobertson, line 238: This should have been a DataError. Otherwise the poller will end up retrying the same sync again and again.

PR #267, opened 2023-08-22
Migration to clean up bogus rooms

### LINKED ISSUE #256: Clean up bad snapshots from existing databases.

DMRobertson:
These rooms have invalid state snapshots, leading to correctness problems. The existence of such snapshots leaves us open to performance problems due to gappy state updates.

We have prevented invalid state snapshots from being formed (#255, #266) in the future. The remaining hole is to discard existing invalid snapshots.

Closes #256.


DMRobertson:
> Can we get some data on how many rooms were affected? It would be nice to have as a sanity check when running this.

I was thinking the same. I guess I will have to write the migration in go for this?

DMRobertson:
I'll need to test this somehow.

state/migrations/20230822180807_bogus_snapshot_cleanup.go
```diff
@@ -0,0 +1,93 @@
...
+		`DELETE FROM `+table+` WHERE room_id = ANY($1)`,
+		pq.StringArray(roomIDs))
+	if err != nil {
+		return fmt.Errorf("failed to delete from %s: %w", table, err)
+	}
+	ra, err := result.RowsAffected()
+	if err != nil {
+		logger.Warn().Err(err).Msgf("Couldn't get number of rows deleted from %s", table)
+	} else {
+		logger.Info().Msgf("Deleted %d snapshots from %s", ra, table)
```
kegsay, line 57:
```suggestion
		logger.Info().Msgf("Deleted %d rows from %s", ra, table)
```
Technically they aren't always snapshots

kegsay (approved):

DMRobertson:
Tested manually in a staging environment and then with a gradual rollout.

PR #269, opened 2023-08-23
OTLP not Jaeger

kegsay:
Replace exporter span format from Jaeger to OTLP as it's more "generic" and thus works with more server-side stuff.

Also add in the option to set HTTP basic auth on the endpoint.


DMRobertson (approved):
Seems sane. Let's try it.

internal/trace.go
```diff
@@ -72,12 +74,34 @@ func StartTask(ctx context.Context, name string) (context.Context, *Task) {
...
-	))
+func ConfigureOTLP(otlpURL, otlpUser, otlpPass, version string) error {
+	ctx := context.Background()
+	parsedOTLPURL, err := url.Parse(otlpURL)
+	if err != nil {
+		return err
+	}
+	isInsecure := parsedOTLPURL.Scheme == "http" // e.g testing and development
+	if parsedOTLPURL.Path != "" {
+		return fmt.Errorf("OTLP URL %s cannot contain any path segments", otlpURL)
```
genofire, line 85: why is a path not allowed?

PR #275, opened 2023-08-25
Try to optimize `LazyLoadTimelines`

S7evinK:
Instead of hitting the database `len(roomIDs) * 2`-times, we're now using lateral joins to get everything in one go.
Executing the query manually with
```sql
WITH input AS (
    select room_id roomID, 0 start, 115389979 end from syncv3_rooms LIMIT 4000
)
```
Yields
```
Planning Time: 0.213 ms
Execution Time: 84.110 ms
```
Compared to getting one room 
```
Planning Time: 0.177 ms
Execution Time: 0.044 ms
```
(naively multiplied by 4000 -> 800ms - this excludes any latency added by calling the DB 4k times over the network)

Once approved, or not, the old method can be removed (just exists to validate the results are the same in the test)


S7evinK:
(closed for being stale)

PR #276, opened 2023-08-29
Increase the number of user IDs allowed in device list changes

kegsay:
This isn't great.

### DIFF
--- a/state/device_data_table.go
+++ b/state/device_data_table.go
@@ -55,7 +55,14 @@ func (t *DeviceDataTable) Select(userID, deviceID string, swap bool) (result *in
 			return err
 		}
 		// unmarshal to swap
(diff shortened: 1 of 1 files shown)


state/device_data_table.go
```diff
@@ -104,7 +104,14 @@ func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (err error) {
 		// unmarshal and combine
 		var tempDD internal.DeviceData
 		if len(row.Data) > 0 {
-			if err = cbor.Unmarshal(row.Data, &tempDD); err != nil {
+			opts := cbor.DecOptions{
```
S7evinK, line 107: We'll need that in the `SELECT` method as well.

S7evinK (approved):

PR #280, opened 2023-08-31
Redact events in the DB on m.room.redaction

### LINKED ISSUE #279: Redact event content when processing m.room.redaction events
I initially reported this in Element X iOS here: https://github.com/vector-im/element-x-ios/issues/1441

Room previews on Element X iOS show the content of the last message, even if that message was redacted before. In an unencrypted room, this also happens on a freshly installed and logged in client, suggesting the proxy is sending the data when it shouldn't.
I first noticed this after upgrading my sliding-sync proxy to v0.99.5 running with the official docker image.

kegsay:
Fixes #279

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+
 	"github.com/matrix-org/sliding-sync/internal"
 
 	"github.com/getsentry/sentry-go"
@@ -398,11 +399,13 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch 
 
 	var latestNID int64
 	newEvents := make([]Event, 0, len(eventIDToNID))
+	var redactTheseEventIDs []string
 	for _, ev := range dedupedEvents {
 		nid, ok := eventIDToNID[ev.ID]
 		if ok {
 			ev.NID = int64(nid)
-			if gjson.GetBytes(ev.JSON, "state_key").Exists() {
+			parsedEv := gjson.ParseBytes(ev.JSON)
+			if parsedEv.Get("state_key").Exists() {
 				// XXX: reusing this to mean "it's a state event" as well as "it's part of the state v2 response"
 				// its important that we don't insert 'ev' at this point as this should be False in the DB.
 				ev.IsState = true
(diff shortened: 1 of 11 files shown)


state/accumulator.go
```diff
@@ -412,11 +415,40 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
 			if ev.NID > latestNID {
 				latestNID = ev.NID
 			}
+			// if this is a redaction, try to redact the referenced event (best effort)
+			if !ev.IsState && ev.Type == "m.room.redaction" {
```
DMRobertson, line 419: Are there any rules about who is allowed to issue redactions? I suppose there might be, but they are the upstream homeserver's responsibility to worry about. (The proxy is behaving like a client here.)

state/accumulator.go
```diff
@@ -412,11 +415,40 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
...
 		}
 	}
 
+	// if we are going to redact things, we need the room version to know the redaction algorithm
+	// so pull it out once now.
+	var roomVersion string
+	if len(redactTheseEventIDs) > 0 {
+		createEventJSON, err := a.eventsTable.SelectCreateEvent(txn, roomID)
+		if err != nil {
+			return 0, nil, fmt.Errorf("SelectCreateEvent: %s", err)
```
DMRobertson, line 440:
```suggestion
			return 0, nil, fmt.Errorf("SelectCreateEvent: %w", err)
```

for error propagation

state/accumulator.go
```diff
@@ -412,11 +415,40 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
...
+		}
+		roomVersion = gjson.GetBytes(createEventJSON, "content.room_version").Str
+		if roomVersion == "" {
+			// Defaults to "1" if the key does not exist.
+			roomVersion = "1"
+		}
+		if err = a.eventsTable.Redact(txn, roomVersion, redactTheseEventIDs); err != nil {
+			return 0, nil, err
+		}
```
DMRobertson, line 449: If something goes wrong here this will stop the accumulation in its tracks. Maybe that's for the best though?

state/event_table.go
```diff
@@ -336,6 +338,30 @@ func (t *EventTable) LatestEventNIDInRooms(txn *sqlx.Tx, roomIDs []string, highe
...
 
+func (t *EventTable) Redact(txn *sqlx.Tx, roomVer string, eventIDs []string) error {
+	// verifyAll=false so if we are asked to redact an event we don't have we don't fall over.
+	eventsToRedact, err := t.SelectByIDs(txn, false, eventIDs)
+	if err != nil {
+		return fmt.Errorf("EventTable.Redact[%v]: %s", eventIDs, err)
+	}
+	rv, err := gomatrixserverlib.GetRoomVersion(gomatrixserverlib.RoomVersion(roomVer))
+	if err != nil {
+		// unknown room version... let's just default to "1"
```
DMRobertson, line 349: Worth a warning? Ditto for the case in the accumulator where you fall back to 1

state/event_table_test.go
```diff
@@ -924,3 +925,168 @@ func TestEventTableSelectUnknownEventIDs(t *testing.T) {
...
+			original: map[string]interface{}{
+				"type":   "m.room.message",
+				"sender": "@nasty-mc-nastyface:localhost",
+				"content": map[string]interface{}{
+					"msgtype": "m.text",
+					"body":    "Something not very nice",
+				},
+			},
+			wantContent: map[string]interface{}{},
+			roomVer:     "10",
```
DMRobertson, line 955: suggest a test case that checks room version 11, since there is explicit logic for dealing with that room ver

DMRobertson (commented):
Looks good. I was expecting that we would have to do some kind of cache invalidation---redacting in-memory copies of events in the global and user caches. Why haven't we done that---do we not keep raw event JSON in memory?

state/accumulator.go
kegsay: It means we will retry as Accumulate is a retryable function. This is good if `err` is because of a temporary DB connection blip. This is bad if it's a data error as it means we'll never make forward progress.

state/accumulator.go
kegsay: Yup.

state/event_table.go
kegsay: This can and will happen for unknown MSC room versions, so yeah warning but not sentry.

state/event_table_test.go
kegsay: Gomatrixserverlib does not support v11.

PR #289, opened 2023-09-07
Add a bit more tracing, temporary Prometheus metric

S7evinK:
Adds some tracing and logging.
Also adds a temporary histogram to get a feeling how often we're creating a new snapshot. This is to see if a change I have in mind (separate membership table) is a viable solution.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -23,13 +23,13 @@ import (
 // Accumulate function for timeline events. v2 sync must be called with a large enough timeline.limit
 // for this to work!
 type Accumulator struct {
-	db              *sqlx.DB
-	roomsTable      *RoomsTable
-	eventsTable     *EventTable
-	snapshotTable   *SnapshotTable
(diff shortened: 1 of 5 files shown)


state/accumulator.go
```diff
@@ -22,12 +23,13 @@ import (
...
-	snapshotTable *SnapshotTable
-	spacesTable   *SpacesTable
-	entityName    string
+	db              *sqlx.DB
+	roomsTable      *RoomsTable
+	eventsTable     *EventTable
+	snapshotTable   *SnapshotTable
+	spacesTable     *SpacesTable
+	entityName      string
+	snapshotSizeVec *prometheus.HistogramVec // TODO: Remove, this is temporary to get a feeling how often a new snapshot is created
```
DMRobertson, line 32: maybe `snapshotMemberCountVec`, since it only counts #members and not the total number of state events?

DMRobertson (approved):

PR #291, opened 2023-09-07
bugfix: set unsigned.redacted_because field on redaction

kegsay:
Element X relies on this field being set.

See https://github.com/matrix-org/matrix-spec/issues/1630 for the correctness of this, but regardless, we need to unbreak clients.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -399,8 +399,8 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch 
 
 	var latestNID int64
 	newEvents := make([]Event, 0, len(eventIDToNID))
-	var redactTheseEventIDs []string
-	for _, ev := range dedupedEvents {
+	redactTheseEventIDs := make(map[string]*Event)
+	for i, ev := range dedupedEvents {
 		nid, ok := eventIDToNID[ev.ID]
 		if ok {
 			ev.NID = int64(nid)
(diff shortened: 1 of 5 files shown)


state/accumulator.go
```diff
@@ -423,7 +423,7 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
 					redactsEventID = parsedEv.Get("content.redacts").Str
 				}
 				if redactsEventID != "" {
-					redactTheseEventIDs = append(redactTheseEventIDs, redactsEventID)
+					redactTheseEventIDs[redactsEventID] = &dedupedEvents[i]
```
DMRobertson, line 426: This is a map from redactee to redaction. It's possible that two people concurrently send redactions for the same event and they both appear in the timeline. Does this matter? I would assume not: redactions are basically idempotent, and we should set `redacted_because` to something so clients will be happy.

state/event_table.go
```diff
@@ -338,7 +338,11 @@ func (t *EventTable) LatestEventNIDInRooms(txn *sqlx.Tx, roomIDs []string, highe
 	return
 }
 
-func (t *EventTable) Redact(txn *sqlx.Tx, roomVer string, eventIDs []string) error {
+func (t *EventTable) Redact(txn *sqlx.Tx, roomVer string, redacteeEventIDToRedactEvent map[string]*Event) error {
+	eventIDs := make([]string, 0, len(redacteeEventIDToRedactEvent))
+	for e := range redacteeEventIDToRedactEvent {
+		eventIDs = append(eventIDs, e)
+	}
```
DMRobertson, line 345: One day we might have https://pkg.go.dev/golang.org/x/exp/maps#Values in the stdlib for this

DMRobertson (approved):

state/accumulator.go
kegsay: Yeah last write wins then.

kegsay:
Manually confirmed the fix.

Issue #294, opened 2023-09-07
Cache invalidation


kegsay:
### Cache invalidation in SS

This is the process of informing downstream API caches that what they have remembered for a room is incorrect and needs refetching from the database. We (dmr & kegan) propose invalidation is room ID scoped, and not more fine-grained for now. This keeps things simple but means we ask the DB for more information than we strictly need to.

Cases when invalidation are needed:
 - Redactions. Whilst the proxy doesn't cache the event content of messages, it definitely does cache things like the room name. If the room name is redacted, we need to inform the global/user/connection caches that the name is now unset. This also applies to member display names (for `room.name`) and avatar URLs (for `room.avatar`), as well as the canonical alias (for `room.name`).
 - State resolution. This is part of a broader strategy for refreshing entire room state snapshots. The downstream room state caches are currently loaded on startup then strictly updated when _live timeline events_ arrive from the poller. This is a problem when:
     * the state changes without a timeline event (state resets),
	 * the proxy has a very old copy of a room which is updated when a poller joins that room.

   In the latter case, we fudged a solution by incorrectly prepending `state` block events to the timeline to get the caches to update downstream correctly. This makes the timeline incorrect when the client asks for a sufficiently high `timeline_limit` though.

We propose to add a new payload type `V2InvalidateRoom` which will contain the room ID to invalidate and the new snapshot ID to load from. This will cause downstream API processes to reload caches from that snapshot. The precise data that needs to be reloaded is the entire list of all joined members and invited members in addition to populating this struct:

```go
type RoomMetadata struct {
	RoomID         string
	Heroes         []Hero
	NameEvent      string // the content of m.room.name, NOT the calculated name
	AvatarEvent    string // the content of m.room.avatar, NOT the resolved avatar
	CanonicalAlias string
	JoinCount      int
	InviteCount    int
	// LastMessageTimestamp is the origin_server_ts of the event most recently seen in
	// this room. Because events arrive at the upstream homeserver out-of-order (and
	// because origin_server_ts is an untrusted event field), this timestamp can
	// _decrease_ as new events come in.
	LastMessageTimestamp uint64
	// LatestEventsByType tracks timing information for the latest event in the room,
	// grouped by event type.
	LatestEventsByType map[string]EventMetadata
	Encrypted          bool
	PredecessorRoomID  *string
	UpgradedRoomID     *string
	RoomType           *string
	// if this room is a space, which rooms are m.space.child state events. This is the same for all users hence is global.
[...]

PR #296, opened 2023-09-07
WIP invalidate cached room data after a redaction

DMRobertson:


state/accumulator.go
```diff
@@ -317,6 +317,16 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
...
 }
 
+type AccumulateResult struct {
+	// TODO: is this redundant---identical to len(TimelineNIDs)?
+	NumNew       int
+	TimelineNIDs []int64
+	// ReloadSnapshot is set to the snapshot ID representing the current state of the room, if this Accumulate has
+	// changed state in a non-incremental fashion. If so, the application should reload its view of this room from the
+	// given snapshot. Otherwise this is 0.
+	ReloadFromSnapshot int64
```
DMRobertson, line 327: This is per https://github.com/matrix-org/sliding-sync/issues/294#issue-1885953239

Not convinced we need this; API could always just load up the latest snapshot?

sync2/handler2/handler.go
```diff
@@ -294,19 +294,27 @@ func (h *Handler) Accumulate(ctx context.Context, userID, deviceID, roomID, prev
...
 	// Insert new events
-	numNew, latestNIDs, err := h.Store.Accumulate(userID, roomID, prevBatch, timeline)
+	accResult, err := h.Store.Accumulate(userID, roomID, prevBatch, timeline)
 	if err != nil {
 		logger.Err(err).Int("timeline", len(timeline)).Str("room", roomID).Msg("V2: failed to accumulate room")
 		internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(err)
 		return err
 	}
 
+	// TODO: does the ordering of this versus V2Accumulate matter here?
```
DMRobertson, line 304: Pretty sure it does. (Want to make sure we redact data as soon as we can)

state/accumulator.go
```diff
@@ -317,6 +317,16 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
 	return res, err
 }
 
+type AccumulateResult struct {
+	// TODO: is this redundant---identical to len(TimelineNIDs)?
```
kegsay, line 321: I don't think so. If you call accumulate 2x with:
 - a,b,c
 - a,b,c,d

Then in the latter case you get timeline nids x4 and num new = 1.

state/accumulator.go
```diff
@@ -494,20 +507,27 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
...
-			return 0, nil, err
+			return AccumulateResult{}, err
 		}
 	}
 
+	if len(redactTheseEventIDs) > 0 {
+		// TODO: this is going to send out a cache invalidation when someone redacts a
+		// message event, which is unnecessary. Can we get eventsTable.Redact to tell us
+		// if it redacted any state events, and only include a snapID here if so?
+		result.RequiresReload = true
```
kegsay, line 518: We probably need to do this before landing this PR as otherwise redactions will be expensive.

state/storage.go
```diff
@@ -306,6 +306,77 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
...
+        WHERE syncv3_rooms.room_id = $1
+    )
+	SELECT event_id, event_type, state_key, event, membership
+	FROM syncv3_events JOIN snapshot ON (
+		event_nid = ANY (ARRAY_CAT(events, membership_events))
+	)
+	WHERE (event_type IN ('m.room.name', 'm.room.avatar', 'm.room.canonical_alias') AND state_key = '')
+	   OR (event_type = 'm.room.member' AND membership IN ('join', '_join', 'invite', '_invite'))
+	ORDER BY event_nid ASC
+	;`, metadata.RoomID)
```
kegsay, line 329: How long does this take on m.org?

sync3/handler/handler.go
```diff
@@ -802,6 +802,15 @@ func (h *SyncLiveHandler) OnExpiredToken(p *pubsub.V2ExpiredToken) {
 	h.ConnMap.CloseConnsForDevice(p.UserID, p.DeviceID)
 }
 
+func (h *SyncLiveHandler) OnInvalidateRoom(p *pubsub.V2InvalidateRoom) {
+	ctx, task := internal.StartTask(context.Background(), "OnInvalidateRoom")
+	defer task.End()
+
+	logger.Warn().Msgf("DMR: invalidate room %s", p.RoomID)
+	h.GlobalCache.OnInvalidateRoom(ctx, p.RoomID)
+	h.Dispatcher.OnInvalidateRoom(ctx, p.RoomID)
```
kegsay, line 811: The dispatcher usually calls the global cache e.g:
```go
	// global listeners (invoke before per-user listeners so caches can update)
	listener := d.userToReceiver[DispatcherAllUsers]
	if listener != nil {
		listener.OnEphemeralEvent(ctx, roomID, ephEvent)
	}
```

sync3/caches/global.go
```diff
@@ -385,3 +385,22 @@ func (c *GlobalCache) OnNewEvent(
...
+	defer c.roomIDToMetadataMu.Unlock()
+
+	metadata, ok := c.roomIDToMetadata[roomID]
+	if !ok {
+		logger.Warn().Str("room_id", roomID).Msg("OnInvalidateRoom: room not in global cache")
+		return
+	}
+
+	logger.Warn().Any("before", metadata).Msgf("DMR: invalidate GLOBALLLLLL room %s", roomID)
+	err := c.store.ResetMetadataState(metadata)
```
kegsay, line 400: Presumably needs to be written back to `c.roomIDToMetadata`?

kegsay (commented):
Looks very sane.

state/accumulator.go
DMRobertson: Hmm. It's a bit surprising to me that we embed all of these events---including those which we know are _not_ new---in the V2Accumulate payload. Is the point that the pubsub might arrive out-of-order though? (E.g. Seeing (d) then (a, b, c) would be bad.)

state/storage.go
DMRobertson: I did some testing on Matrix HQ---let me rerun and get you some numbers.

sync3/caches/global.go
DMRobertson: Here `metadata` is a pointer and I'm updating `*metadata` in-place. We have a lock on roomIDToMetadataMu in the caller so this might be safe. I wasn't completely sure though, see e.g. https://github.com/matrix-org/sliding-sync/blob/1b14ac3d1f8b828b378acf854c9ff350339f2895/state/storage.go#L311-L312

state/storage.go
DMRobertson: According to notes this was ~200ms for Matrix HQ. Running it now is ~270ms.

```
postgres=> \set room_id !OGEhHVWSdvArJzumhm:matrix.org \set snapshot_id 682264 \\
postgres=> 
postgres=> EXPLAIN ANALYZE
postgres-> WITH snapshot(events, membership_events) AS (
postgres(>     SELECT events, membership_events
postgres(>     FROM syncv3_snapshots
postgres(>         JOIN syncv3_rooms ON snapshot_id = current_snapshot_id
postgres(>     WHERE syncv3_rooms.room_id = :'room_id'
postgres(> )
postgres-> SELECT event_id, event_type, state_key, event, membership
postgres-> FROM syncv3_events JOIN snapshot ON (
postgres(>         event_nid = ANY (ARRAY_CAT(events, membership_events))
postgres(>     )
postgres-> WHERE (event_type IN ('m.room.name', 'm.room.avatar', 'm.room.canonical_alias') AND state_key = '')
postgres->    OR (event_type = 'm.room.member' AND membership IN ('join', '_join', 'invite', '_invite'))
postgres-> ORDER BY event_nid ASC
postgres-> ;
                                                                                                                   QUERY PLAN                 
                                                                                                   
----------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------
 Sort  (cost=24.28..24.28 rows=2 width=501) (actual time=246.204..258.618 rows=48046 loops=1)
   Sort Key: syncv3_events.event_nid
   Sort Method: external merge  Disk: 19976kB
   ->  Nested Loop  (cost=1.15..24.27 rows=2 width=501) (actual time=17.091..200.713 rows=48046 loops=1)
         ->  Nested Loop  (cost=0.71..16.75 rows=1 width=271) (actual time=0.032..0.034 rows=1 loops=1)
               ->  Index Scan using syncv3_rooms_pkey on syncv3_rooms  (cost=0.29..8.30 rows=1 width=8) (actual time=0.020..0.021 rows=1 loops
=1)
                     Index Cond: (room_id = '!OGEhHVWSdvArJzumhm:matrix.org'::text)
               ->  Index Scan using syncv3_snapshots_snapshot_id_room_id_key on syncv3_snapshots  (cost=0.42..8.44 rows=1 width=279) (actual t
ime=0.010..0.010 rows=1 loops=1)
                     Index Cond: (snapshot_id = syncv3_rooms.current_snapshot_id)
         ->  Index Scan using syncv3_events_pkey on syncv3_events  (cost=0.44..7.50 rows=2 width=501) (actual time=15.542..192.561 rows=48046 
loops=1)
               Index Cond: (event_nid = ANY (array_cat(syncv3_snapshots.events, syncv3_snapshots.membership_events)))
               Filter: (((event_type = ANY ('{m.room.name,m.room.avatar,m.room.canonical_alias}'::text[])) AND (state_key = ''::text)) OR ((ev
ent_type = 'm.room.member'::text) AND (membership = ANY ('{join,_join,invite,_invite}'::text[]))))
               Rows Removed by Filter: 53701
 Planning Time: 2.699 ms
 Execution Time: 268.729 ms
(15 rows)
```

state/storage.go
```diff
@@ -306,6 +306,77 @@ func (s *Storage) MetadataForAllRooms(txn *sqlx.Tx, tempTableName string, result
...
+	if err != nil {
+		return fmt.Errorf("ResetMetadataState[%s]: %w", metadata.RoomID, err)
+	}
+
+	heroMemberships := circularSlice[*Event]{max: 6}
```
DMRobertson, line 334: Note: when populating the heroes I only care about the 6 most recent heroes. However to get accurate join and invite counts we effectively need to pull all the members out.

sync3/handler/handler.go
DMRobertson: 6f37ac0

state/accumulator.go
```diff
@@ -494,20 +510,40 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
...
+		redactedEventIDs := make([]string, 0, len(redactTheseEventIDs))
+		for eventID := range redactTheseEventIDs {
+			redactedEventIDs = append(redactedEventIDs, eventID)
+		}
+		var currentStateRedactions int
+		err = a.db.Get(&currentStateRedactions, `
+			SELECT COUNT(*)
+			FROM syncv3_events
+			    JOIN syncv3_snapshots ON event_nid = ANY (ARRAY_CAT(events, membership_events))
+			WHERE snapshot_id = $1 AND event_id = ANY($2)
```
DMRobertson, line 529: Sanity checked the query plan as follows:

```
\set room_id !OGEhHVWSdvArJzumhm:matrix.org  \\
\set snapshotID 756886 \\
\set event1 $liLBpCuSUVPqPVsff3jCFT9VN9dW551TiDFx5QE8MmI \\
\set event2 $ARQ3jrwiMsX1pLeDIjvOW_ga4CucgAnghrzHWwlULS0 \\

EXPLAIN ANALYZE SELECT COUNT(*)
FROM syncv3_events
    JOIN syncv3_snapshots ON event_nid = ANY (ARRAY_CAT(events, membership_events))
WHERE snapshot_id = :snapshotID AND event_id = ANY (ARRAY[:'event1', :'event2'])
;

                                                                               QUERY PLAN                                                     
                           
----------------------------------------------------------------------------------------------------------------------------------------------
---------------------------
 Aggregate  (cost=25.64..25.65 rows=1 width=8) (actual time=3.044..3.045 rows=1 loops=1)
   ->  Nested Loop  (cost=0.99..25.64 rows=1 width=0) (actual time=1.495..3.040 rows=2 loops=1)
         Join Filter: (syncv3_events.event_nid = ANY (array_cat(syncv3_snapshots.events, syncv3_snapshots.membership_events)))
         ->  Index Scan using syncv3_snapshots_snapshot_id_room_id_key on syncv3_snapshots  (cost=0.42..8.44 rows=1 width=271) (actual time=0.
024..0.025 rows=1 loops=1)
               Index Cond: (snapshot_id = 756886)
         ->  Index Scan using syncv3_events_event_id_key on syncv3_events  (cost=0.56..17.15 rows=2 width=8) (actual time=0.024..0.038 rows=2 
loops=1)
               Index Cond: (event_id = ANY ('{$liLBpCuSUVPqPVsff3jCFT9VN9dW551TiDFx5QE8MmI,$ARQ3jrwiMsX1pLeDIjvOW_ga4CucgAnghrzHWwlULS0}'::tex
t[]))
 Planning Time: 0.577 ms
 Execution Time: 3.073 ms
(9 rows)
```

state/storage.go
DMRobertson: If we're only worrying about redactions then there aren't gonna be any count changes. But as #294 says there's the more general problem of "our state is completely out of date".

state/storage.go
DMRobertson: Eyeballing it, the ORDER BY adds ~70ms, which surprised me! (I want to process the Nids in ascending order to pick out the heroes.)

DMRobertson: I wonder if it might be quicker to do the sorting in Go? Suggest this is good enough to leave as-is for now though.

(Also I'm surprised that we use the 6 newest heroes, not the 6 oldest)

state/accumulator.go
DMRobertson: Have done this. Added a unit test to keep me honest, too.

state/accumulator.go
DMRobertson: > Then in the latter case you get timeline nids x4 and num new = 1.

I think this is wrong.

- The second Accumulate call would return deduped events `[d]` thanks to this logic: https://github.com/matrix-org/sliding-sync/blob/6f3f556842a5e04c3ce67aec4204292f3e55c9c3/state/accumulator.go#L598 
- numNew should be the same as len(dedupedEvents) due to this: https://github.com/matrix-org/sliding-sync/blob/main/state/accumulator.go#L396-L404
- timlineNIDs corresponds to dedupedEvents thanks to this loop: https://github.com/matrix-org/sliding-sync/blob/main/state/accumulator.go#L409-L438

kegsay (approved):
Excellent

sync3/caches/global.go
DMRobertson: From discussion elsewhere:

> it should be safe so long as you're doing that from the subscriber goroutine, and not per-conn goroutines

Issue #250, opened 2023-08-14
Duplicate leave events


kegsay:
I suspect since https://github.com/matrix-org/sliding-sync/pull/235 we now send N leave events, once per poller as this is now tied to `OnLeftRoom` and isn't de-duped as we explicitly ignore the load position.

Reported by Ivan, running `f73c8e4604d6`

S7evinK:
This should now be fixed by https://github.com/matrix-org/sliding-sync/pull/297

PR #297, opened 2023-09-08
Fix duplicate leave events

S7evinK:
Still investigating, but doesn't even seem to require multiple pollers as per https://github.com/matrix-org/sliding-sync/issues/250

### DIFF
--- a/tests-e2e/membership_transitions_test.go
+++ b/tests-e2e/membership_transitions_test.go
@@ -466,6 +466,72 @@ func TestLeavingRoomReturnsOneEvent(t *testing.T) {
 	}
 }
 
+// Basically the same as above, without joining the room (separate test to make sure
+// we don't already have a poller running for Alice)
+func TestRejectingInviteReturnsOneEvent(t *testing.T) {
+	alice := registerNewUser(t)
+	bob := registerNewUser(t)
+	roomName := "It's-a-me-invitio"
+
+	for _, aliceSyncing := range []bool{false, true} {
+		t.Run(fmt.Sprintf("rejecting an invite returns one leave event (multiple poller=%v)", aliceSyncing), func(t *testing.T) {
+			inviteRoomID := alice.CreateRoom(t, map[string]interface{}{"preset": "private_chat", "name": roomName})
+			t.Logf("TestRejectingInviteReturnsOneEvent room %s", inviteRoomID)
+
+			if aliceSyncing {
+				alice.SlidingSync(t, sync3.Request{})
+			}
+
+			// sync as bob, we see no invites yet.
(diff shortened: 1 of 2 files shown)


tests-e2e/membership_transitions_test.go
```diff
@@ -397,6 +398,140 @@ func TestInviteRejectionTwice(t *testing.T) {
...
+			bob.JoinRoom(t, inviteRoomID, []string{})
+			res = bob.SlidingSyncUntilMembership(t, res.Pos, inviteRoomID, bob, "join")
+			t.Logf("bob joined")
+
+			// leave the room again, we should receive exactly one leave response
+			bob.LeaveRoom(t, inviteRoomID)
+			res = bob.SlidingSyncUntilMembership(t, res.Pos, inviteRoomID, bob, "leave")
+
+			if room, ok := res.Rooms[inviteRoomID]; ok {
+				// If alice is NOT syncing, we run into this failure mode
```
kegsay, line 444: Can you confirm what part of this test was failing without the diff to `user.go`?

tests-e2e/membership_transitions_test.go
```diff
@@ -397,6 +398,140 @@ func TestInviteRejectionTwice(t *testing.T) {
...
+					t.Errorf("[multiple poller=%v] expected 1 timeline event, got %d", aliceSyncing, c)
+				}
+			} else {
+				t.Errorf("[multiple poller=%v] expected room %s in response, but didn't find it", aliceSyncing, inviteRoomID)
+			}
+
+			res = bob.SlidingSync(t, sync3.Request{}, WithPos(res.Pos))
+
+			// this should not happen, as we already send down the leave event
+			// If alice is synching, we run into this failure mode
```
kegsay, line 458:
```suggestion
			// If alice is syncing, we run into this failure mode
```

tests-e2e/membership_transitions_test.go
```diff
@@ -397,6 +398,140 @@ func TestInviteRejectionTwice(t *testing.T) {
...
```
kegsay, line 524:
```suggestion
			// If alice is syncing, we run into this failure mode
```

kegsay (approved):

tests-e2e/membership_transitions_test.go
S7evinK: If Alice is _NOT_ syncing, we receive the leave event twice, once through  `Accumulate` and once through `OnLeftRoom`. 

If Alice _IS_ syncing however, the test fails for exactly the same reason - we receive the same event twice, but at a different place in the test. We're syncing as Bob until we left the room, after that we do another `SlidingSync` call, which again receives the event.

PR #298, opened 2023-09-12
Reuse existing connections

### LINKED ISSUE #274: OIDC expirations cause more frequent expired connections
When the poller receives a 401, it calls `OnExpiredToken` which removes the token (not device) from the DB then pokes pubsub with `pubsub.V2ExpiredToken` and the user/device ID affected. This then hits this code:
```
func (h *SyncLiveHandler) OnExpiredToken(p *pubsub.V2ExpiredToken) {
	h.EnsurePoller.OnExpiredToken(p)
	h.ConnMap.CloseConnsForDevice(p.UserID, p.DeviceID)
}
```
Which:
 - Removes the user/device from the ensure polling map, meaning the next time we see a request from this user we need to block on a sync v2 request.
 - Close all SS connections for this user/device.

The latter is troublesome, because it means clients will see M_UNKNOWN_POS on their next request. This then causes the client to do an initial sync again. To provide a seamless transition between tokens, it would be better if the proxy:
 - kept the SS connections open `OnExpiredToken`, but flagged it as expired in some way.
 - when receiving the next client request:
    * if the connect is not expired, continue as normal.
    * if the connection is expired:
        - if the new token is valid (that is, /whoami returns 200 OK) then unexpire the connection and continue, without blocking on ensure polling.
        - else expire the connection and return M_UNKNOWN_TOKEN.

So in other words, instead of eagerly expiring connections when the poller gets a 401, instead lazily expire connections when the next client request gets M_UNKNOWN_TOKEN. In the case of a logout, this next request will never arrive, but the 30min TTL will clean up the connection.

S7evinK:
This should fix #274

Instead of closing all connections, we now set a `expired` flag. This, in combination with a `pos` in the request, is used to reuse existing connections.
If the client doesn't come back for 30min, the connection is expired as usual. (see [comment](https://github.com/matrix-org/sliding-sync/issues/274#issuecomment-1710347743))

### DIFF
--- a/sync3/conn.go
+++ b/sync3/conn.go
@@ -62,6 +62,9 @@ type Conn struct {
 	mu                         *sync.Mutex
 	cancelOutstandingRequest   func()
 	cancelOutstandingRequestMu *sync.Mutex
+
+	// If we think this session is expired, e.g. due to the poller receiving HTTP 401
+	expired bool
 }
 
 func NewConn(connID ConnID, h ConnHandler) *Conn {
@@ -75,6 +78,10 @@ func NewConn(connID ConnID, h ConnHandler) *Conn {
 
 func (c *Conn) Alive() bool {
 	return c.handler.Alive()
+}
+
+func (c *Conn) Expired() bool {
+	return c.expired
 }
 
 func (c *Conn) OnUpdate(ctx context.Context, update caches.Update) {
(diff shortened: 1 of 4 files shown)


kegsay:
Closed as stale. I think we already do something almost like this now anyway.

Issue #295, opened 2023-09-07
Failing to initialise snapshots due to bad data should not cause poller retry loops


kegsay:
They currently do, meaning the poller is effectively wedged and never makes forward progress. "Bad data" in an initialise snapshot means there is no `m.room.create` event in it.

We _think_ this should be impossible for new deployments, and this is mostly a historical artifact when we were more lax in accepting what Synapse was telling us in `/sync`. There is no evidence this affects any other deployment than matrix.org currently.

To fix this, we should return an `*internal.DataError` so the poller skips over this bad data.

PR #299, opened 2023-09-12
bugfix: don't wedge pollers when they get bad state blocks

kegsay:
We returned an error in Initialise when there is no create event
in the state block. If this happens, the poller automatically retries
the same request, tightlooping as the data is always going to cause
the same error. Instead, return an `internal.DataError` which is a way
of expressly telling the pollers to continue and advance the since token.

With regression test. Fixes https://github.com/matrix-org/sliding-sync/issues/295 

NB: This regression test also caught a potential bug which this PR could
have introduced. This PR lets the since token skip over bad responses, but
we want to make damn sure we process EVERYTHING ELSE in that response. In
particular, we could have skipped over large sections of `parseRoomsResponse`
as we would bail early on errors. This is now fixed to bail only after processing
the entire joined rooms map.

There's additional checks and tests to make sure retriable errors are prioritised over unretriable errors.


state/accumulator.go
```diff
@@ -238,7 +238,7 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
 				Str("room_id", roomID).
 				Int("len_state", len(events)).
 				Msg(errMsg)
-			return fmt.Errorf(errMsg)
+			return internal.NewDataError(errMsg)
```
DMRobertson, line 241: To check: this means "the homeserver has given us duff data and there's no point retrying this"?

DMRobertson: Might be worth a comment to this effect as on +392 below

tests-integration/regressions_test.go
```diff
@@ -228,3 +229,119 @@ func TestMalformedEventsState(t *testing.T) {
...
+	defer v3.close()
+
+	goodRoom := "!good:localhost"
+	badRoom := "!bad:localhost"
+
+	v2.addAccount(t, alice, aliceToken)
+	// we should see the since token increment, if we see repeats it means
+	// we aren't returning DataErrors when we should be.
+	wantSinces := []string{"", "1", "2"}
+	ch := make(chan bool)
```
DMRobertson, line 255: To check: we don't use the bool here; we're just using the channel a something that we can `select` over and close to indicate completion?

.github/workflows/tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             - name: Install Go
               uses: actions/setup-go@v4
               with:
-                  go-version: 1.19
+                  go-version: 1.20
```
DMRobertson, line 48: This is a yaml literal float equal to `1.2`---probably need to quote this.

DMRobertson (commented):
Happy if CI is.

tests-integration/regressions_test.go
kegsay: Yes

.github/workflows/tests.yml
kegsay: Oh fun.

kegsay:
Putting this on hold as more pressing fires but so I don't forget:
 - need to fix CI
 - my error handling breaks if we have to-device events that have a retriable error as I only did the error handling for `parseRoomsResponse` - it needs to be for all errors.

kegsay:
> my error handling breaks if we have to-device events that have a retriable error as I only did the error handling for parseRoomsResponse - it needs to be for all errors.

Failed to repro my concerns here, which were if you got 1x response with BOTH to-device and room data. If we returned a `DataError` (indicating plz increment since) and then hit a retriable error on to-device, that would we increment since regardless. This isn't how the code is structured though:
```go
	retryErr = p.parseRoomsResponse(ctx, resp)
	if shouldRetry(retryErr) {
		p.logger.Err(retryErr).Msg("Poller: parseRoomsResponse returned an error")
		s.failCount += 1
		return nil
	}
	// process to-device messages as the LAST retryable data so we don't double-process
	// to-device msgs on retrys. In other words, if parseToDeviceMessages returns no error
	// then we for sure are going to increment the since token, so cannot see duplicates.
	// If parseToDeviceMessages was earlier, a later parse function could force a retry,
	// causing duplicates. Other parse functions don't have this problem as they are
	// deduplicated.
	retryErr = p.parseToDeviceMessages(ctx, resp)
	if shouldRetry(retryErr) {
		p.logger.Err(retryErr).Msg("Poller: parseToDeviceMessages returned an error")
		s.failCount += 1
		return nil
	}
```

So:
 - if we should retry on a room  response, we will not increment since and hence see the to-device msgs again.
 - if we should NOT retry on a room (e.g bad state block), we continue processing and call `parseToDeviceMessages`.
 - If that then hits a retriable error, we retry and don't increment since.

state/accumulator.go
```diff
@@ -389,6 +390,7 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, prevBatch
 				})
 				sentry.CaptureMessage(msg)
 			})
+			// the HS gave us bad data so there's no point retrying
 			// by not returning an error, we are telling the poller it is fine to not retry this request.
 			return 0, nil, nil
```
DMRobertson: I wonder if this ought to be a data error too?

DMRobertson (approved):

state/accumulator.go
kegsay: In both cases we log and continue. Unsure either way. Inertia says leaving this.

Issue #283, opened 2023-09-01
OIDC expirations cause incorrect timelines


kegsay:
because of the gappy syncs, and how initial syncs use timeline limit: 1 to pull in valid prev batch tokens, it means timeline_limit: 20 is pulling in the actual latest event then prepend state events then whatever crap was in the timeline before the last poller expired, see picture.

<img width="943" alt="Screenshot 2023-09-01 at 14 11 57" src="https://github.com/matrix-org/sliding-sync/assets/7190048/911d585c-ea06-415e-8fdf-0d6990ef7ead">

Ideally we would just return the latest event in this scenario, with a correct prev batch token so the client can backfill the data but that means we need to know where the gaps are, so we can then terminate the timeline fetch at the right spot.

The gap flag is a property of the event in the events table, as that table _is_ the timeline ordering.

kegsay:
Strictly not limited to OIDC:

> i.e A and B in room X, only A using the proxy, send 20 msgs then sign out, some msgs go by, some state events, then B uses the proxy -> prepend the missed state+ see latest event only, then request timeline_limit: 20 and you'll see the same thing.

PR #300, opened 2023-09-13
Don't load events when there's a gap between known events

DMRobertson:
Fixes https://github.com/matrix-org/sliding-sync/issues/283
Part of #294.
Is going to conflict with #296.

Recommend commitwise review. I was going to write an E2E test, but it's fiddley to get a gappy sync deterministically, so opted for an integration test.


state/event_table.go
```diff
@@ -280,7 +289,7 @@ func (t *EventTable) SelectUnknownEventIDs(txn *sqlx.Tx, maybeUnknownEventIDs []
 	// array entries. But I don't think that's guaranteed. Return an (unordered) set
 	// out of paranoia.
 	queryStr := `
-	WITH maybe_unknown_events(event_id) AS (SELECT unnest($1::text[]))
+	WITH maybe_unknown_events(event_id) AS (SELECT UNNEST($1::text[]))
```
DMRobertson, line 292: Sorry, this is goland autoformatting the SQL. I haven't worked out how to stop it from doing this yet.

state/migrations/20230913120537_events_missing_previous.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
+ALTER TABLE IF EXISTS syncv3_events
+    ADD COLUMN IF NOT EXISTS missing_previous BOOLEAN NOT NULL DEFAULT FALSE;
```
DMRobertson, line 3: Not sure if this is the right thing to do. Considerations:

- I'd prefer to avoid a nullable column.
- Adding a column with a default should be fast [these days](https://www.postgresql.org/docs/current/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN).
- Don't want the result of the migration from being different to the "proper" schema in the events table source code.
- I could make the migration drop the default... but that would presumably rewrite the entire table.

Also:
- Not sure the default value is correct here. Arguably TRUE would be more correct (we don't know if these events came from limited syncs; maybe we ought to assume the worst)

kegsay: Default FALSE makes sense to me as historically we always assume we have the previous.

internal/types.go
```diff
@@ -8,3 +10,9 @@ type Receipt struct {
 	ThreadID  string `db:"thread_id"`
 	IsPrivate bool
 }
+
+type TimelineResponse struct {
```
kegsay, line 14: Isn't this just a `sync2.TimelineResponse`? Why not use that?

state/accumulator.go
```diff
@@ -333,15 +333,32 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, timeline i
...
 
+	// If this timeline was limited and we don't recognise its first event E3, mark it
+	// as not knowing its previous timeline event.
+	//
+	// We _may_ actually know the previous event. E.g. if the previous poll ends in E1
+	// and we see a limited sync starting with E3, the "gap" may have already been
+	// filled in by other pollers.
+	//
+	//  A:  E1 persisted, E2 omitted and unknown, E3 unknown       (missing previous)
+	//  B:  E1 persisted, E2 omitted and   known, E3 unknown   (not missing previous)
```
kegsay, line 354: Terminology is too confusing. Hard to grok what:
 - omitted
 - unknown
 - missing previous

are all supposed to mean. The proxy has either seen the event or not, why do we have 3^2 states?

state/accumulator_test.go
```diff
@@ -618,6 +619,140 @@ func TestAccumulatorConcurrency(t *testing.T) {
...
+		t.Fatalf("failed to Initialise accumulator: %s", err)
+	}
+
+	steps := []struct {
+		Desc            string
+		Events          []json.RawMessage
+		Limited         bool
+		NumNew          int
+		CheckDesc       string
+		MissingPrevious map[string]bool
```
kegsay, line 646: Please be clearer on the inputs/outputs, otherwise it's hard to see what is case data and what is assertion output.

state/event_table_test.go
```diff
@@ -1127,3 +1127,90 @@ func TestEventTableRedactMissingOK(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to start txn: %s", err)
+	}
+	defer txn.Rollback()
+	roomID := fmt.Sprintf("!%s", t.Name())
+	events := []Event{
+		{ID: "A", MissingPrevious: false},
+		{ID: "B", MissingPrevious: false},
+		{ID: "C", MissingPrevious: false},
+		{ID: "D", MissingPrevious: true},
```
kegsay, line 1145: Be kinder and mux in "missing previous" to the ID please, else I need to remember it's D,G,H.

state/event_table_test.go
```diff
@@ -1127,3 +1127,90 @@ func TestEventTableRedactMissingOK(t *testing.T) {
...
+		},
+		{
+			FromIDExclusive: "E",
+			ToIDInclusive:   "J",
+			ExpectIDs:       []string{"J", "I", "H"},
+		},
+		{
+			FromIDExclusive: "B",
+			ToIDInclusive:   "I",
+			ExpectIDs:       []string{"I", "H"},
```
kegsay, line 1201: Needs more prose explaining what you're testing in each case and why they are expected.

kegsay (commented):
Overall methodology looks great.

state/event_table.go
```diff
@@ -125,8 +125,13 @@ func NewEventTable(db *sqlx.DB) *EventTable {
 		prev_batch TEXT,
 		membership TEXT,
 		is_state BOOLEAN NOT NULL, -- is this event part of the v2 state response?
-		event BYTEA NOT NULL
+		event BYTEA NOT NULL,
+		missing_previous BOOLEAN NOT NULL DEFAULT FALSE
 	);
+
+	COMMENT ON COLUMN syncv3_events.missing_previous IS
+    'True iff the previous timeline event is not known to the proxy.';
```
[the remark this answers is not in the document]
DMRobertson: `COMMENT ON` makes it part of the postgres schema. In psql it'll show up if you describe the table with `\d+`.  For example, from Synapse:


```
matrix=> \d+ destinations
                                                                                                             Table "matrix.destinations"
             Column              │  Type  │ Collation │ Nullable │ Default │ Storage  │ Compression │ Stats target │                                                           Description                                                           
═════════════════════════════════╪════════╪═══════════╪══════════╪═════════╪══════════╪═════════════╪══════════════╪═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 destination                     │ text   │           │ not null │         │ extended │             │              │ server name of remote homeserver in question
 retry_last_ts                   │ bigint │           │          │         │ plain    │             │              │ The last time we tried and failed to reach the remote server, in ms.                                                           ↵
                                 │        │           │          │         │          │             │              │ This field is reset to `0` when we succeed in connecting again.
 retry_interval                  │ bigint │           │          │         │ plain    │             │              │ How long, in milliseconds, to wait since the last time we tried to reach the remote server before trying again.                ↵
                                 │        │           │          │         │          │             │              │ This field is reset to `0` when we succeed in connecting again.
 failure_ts                      │ bigint │           │          │         │ plain    │             │              │ The first time we tried and failed to reach the remote server, in ms.                                                          ↵
                                 │        │           │          │         │          │             │              │ This field is reset to `NULL` when we succeed in connecting again.
 last_successful_stream_ordering │ bigint │           │          │         │ plain    │             │              │ Stream ordering of the most recently successfully sent PDU to this server, sent through normal send (not e.g. backfill).       ↵
                                 │        │           │          │         │          │             │              │ In Catch-Up Mode, the original PDU persisted by us is represented here, even if we sent a later forward extremity in its stead.↵
                                 │        │           │          │         │          │             │              │ See `destination_rooms` for more information about catch-up.
```

internal/types.go
DMRobertson: Yes. There was some kind of import cycle when I tried to use this in more places. Let me see if I can dig the details up.

DMRobertson: If I move this back to sync2 and then try to build sync2 tests, I get a failure to build test binaries:

```
$ go list ./... | xargs -n1 go test -c -o /dev/null
?   	github.com/matrix-org/sliding-sync	[no test files]
?   	github.com/matrix-org/sliding-sync/cmd/syncv3	[no test files]
?   	github.com/matrix-org/sliding-sync/cmd/testmigration	[no test files]
?   	github.com/matrix-org/sliding-sync/pubsub	[no test files]
?   	github.com/matrix-org/sliding-sync/sqlutil	[no test files]
# github.com/matrix-org/sliding-sync/sync2
package github.com/matrix-org/sliding-sync/sync2
	imports github.com/matrix-org/sliding-sync/state: import cycle not allowed in test
FAIL	github.com/matrix-org/sliding-sync/sync2 [setup failed]
FAIL
?   	github.com/matrix-org/sliding-sync/testutils	[no test files]
?   	github.com/matrix-org/sliding-sync/testutils/m	[no test files]
```

Dep graph on main (`6f3f556`)

```
godepgraph --onlyprefixes github.com/matrix-org/sliding-sync github.com/matrix-org/sliding-sync | sed -e 's|github.com/matrix-org/sliding-sync/\?||g' | dot -Tpng -o godepgraph.png
```

godepgraph

I don't understand where I've broken this. On `main` there _are_ references to sync2 inside the `state` tests, and to `state` inside the `sync2` code.


```
godepgraph -withtests -onlyprefixes github.com/matrix-org/sliding-sync github.com/matrix-org/sliding-sync | sed -e 's|github.com/matrix-org/sliding-sync/\?||g' | dot -Tpng -o godepgraph.png
```

godepgraph

DMRobertson: Removing the root package and `internal`:

```
godepgraph -onlyprefixes github.com/matrix-org/sliding-sync -i github.com/matrix-org/sliding-sync/internal github.com/matrix-org/sliding-sync | sed -e '/"github.com\/matrix-org\/sliding-sync"/d' -e 's|github.com/matrix-org/sliding-sync/||g' | dot -Tpng -o godepgraph.png
```

godepgraph

DMRobertson: And even more readable:

```
godepgraph -onlyprefixes github.com/matrix-org/sliding-sync -i github.com/matrix-org/sliding-sync/internal github.com/matrix-org/sliding-sync | sed -e '/"github.com\/matrix-org\/sliding-sync"/d' -e 's|github.com/matrix-org/sliding-sync/||g' -e '/splines/d | dot -Tpng -o godepgraph.png
```

godepgraph

DMRobertson: The cycle was removed in #309.

state/accumulator.go
```diff
@@ -333,15 +333,32 @@ func (a *Accumulator) Accumulate(txn *sqlx.Tx, userID, roomID string, timeline i
 	// - there to be no duplicate events
 	// - if there are new events, they are always new.
 	// Both of these assumptions can be false for different reasons
-	dedupedEvents, err := a.filterAndParseTimelineEvents(txn, roomID, timeline.Events, timeline.PrevBatch)
+	dedupedEvents := parseAndDeduplicateTimelineEvents(roomID, timeline)
+	filteredEvents, err := a.filterToNewTimelineEvents(txn, dedupedEvents)
```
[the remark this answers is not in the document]
DMRobertson: Agreed. ed83ab9

state/accumulator.go
DMRobertson: a5c02b2 tries to explain this better.

DMRobertson:
Upgrade test is failing. After the upgrade:

```
Sync v3 [0.99.10] (262f8cb)
Debug=true LogLevel= MaxConns=0
panic: pq: column "missing_previous" of relation "syncv3_events" does not exist

goroutine 1 [running]:
github.com/jmoiron/sqlx.MustExec(...)
	/home/runner/go/pkg/mod/github.com/jmoiron/sqlx@v1.3.3/sqlx.go:722
github.com/jmoiron/sqlx.(*DB).MustExec(0x30?, {0xdbf993?, 0xc0003c4c60?}, {0x0?, 0xc0002ef900?, 0x40fd47?})
	/home/runner/go/pkg/mod/github.com/jmoiron/sqlx@v1.3.3/sqlx.go:366 +0x46
github.com/matrix-org/sliding-sync/state.NewEventTable(...)
	/home/runner/work/sliding-sync/sliding-sync/state/event_table.go:116
github.com/matrix-org/sliding-sync/state.NewStorageWithDB(0xc0003c4c90, 0x0)
	/home/runner/work/sliding-sync/sliding-sync/state/storage.go:85 +0x70
github.com/matrix-org/sliding-sync.Setup({0xc00004248e?, 0xd7fc8b?}, {0xc00004806a, 0x4c}, {0xc00003c00e, 0xa}, {0x0, 0x0, 0x0, 0x3b9aca00, ...})
	/home/runner/work/sliding-sync/sliding-sync/v3.go:106 +0x189
main.main()
	/home/runner/work/sliding-sync/sliding-sync/cmd/syncv3/main.go:213 +0x175b
```

DMRobertson:
> That's a surprise to me because I thought CREATE TABLE IF NOT EXISTS would be a no-op if the table already exists!

Oh this will be the COMMENT ON..., won't it?

state/event_table.go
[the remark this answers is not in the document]
DMRobertson: As much as I would like to have this in-schema, this is going to fail after an upgrade because we run NewEventTable before the migrations, and `COMMENT ON COLUMN` has no `IF NOT EXISTS` option. 

Going to drop this for now.

sync2/client.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/matrix-org/sliding-sync/internal"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
```
kegsay, line 7: Why has this appeared?

state/event_table.go
```diff
@@ -128,12 +128,10 @@ func NewEventTable(db *sqlx.DB) *EventTable {
 		membership TEXT,
 		is_state BOOLEAN NOT NULL, -- is this event part of the v2 state response?
 		event BYTEA NOT NULL,
+		-- True iff the previous timeline event is not known to the proxy.
```
kegsay, line 131: Definitely add some prose about the fact that we check `limited`, otherwise in the common case of 1 single new event in a sync v2 response, do we know the previous event?

state/event_table.go
```diff
@@ -319,7 +317,7 @@ func (t *EventTable) LatestEventInRooms(txn *sqlx.Tx, roomIDs []string, highestN
 	err = txn.Select(
 		&events,
 		`SELECT event_nid, room_id, event_replaces_nid, before_state_snapshot_id, event_type, state_key, event FROM syncv3_events
-		WHERE event_nid IN (SELECT max(event_nid) FROM syncv3_events WHERE event_nid <= $1 AND room_id = ANY($2) GROUP BY room_id)`,
+		WHERE event_nid IN (SELECT MAX(event_nid) FROM syncv3_events WHERE event_nid <= $1 AND room_id = ANY($2) GROUP BY room_id)`,
```
kegsay, line 320: I'm guessing this is some linter doing this, or just your OCD? ;)

state/event_table_test.go
```diff
@@ -1165,52 +1166,59 @@ func TestEventTable_SelectLatestEventsBetween_MissingPrevious(t *testing.T) {
...
+			Desc:            "covers three gaps, the latter two consecutive - should only return from the second gap onwards",
 			FromIDExclusive: "B",
 			ToIDInclusive:   "I",
-			ExpectIDs:       []string{"I", "H"},
+			ExpectIDs:       []string{"I", "H-gap"},
 		},
 	}
 
 	for _, tc := range testcases {
+		// We're using the notation (X, Y] for a half-open interval excluding X but including Y.
```
kegsay, line 1213: I personally prefer [rust's notation](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html) of `X..=Y`

state/event_table_test.go
```diff
@@ -1138,15 +1138,16 @@ func TestEventTable_SelectLatestEventsBetween_MissingPrevious(t *testing.T) {
...
 		{ID: "B", MissingPrevious: false},
 		{ID: "C", MissingPrevious: false},
-		{ID: "D", MissingPrevious: true},
+		{ID: "D-gap", MissingPrevious: true},
 		{ID: "E", MissingPrevious: false},
 		{ID: "F", MissingPrevious: false},
-		{ID: "G", MissingPrevious: true},
-		{ID: "H", MissingPrevious: true},
+		{ID: "G-gap", MissingPrevious: true},
+		{ID: "H-gap", MissingPrevious: true},
```
kegsay, line 1150: I think I'm just having a lot of mental block here because I see A B C D and instinctively think there is no gap. When you then put `H-gap` I think "uhh before or after"? Can we +2 letters to hint at the gap maybe? E.g A B C F-gap?

state/event_table_test.go
```diff
@@ -1165,52 +1166,59 @@ func TestEventTable_SelectLatestEventsBetween_MissingPrevious(t *testing.T) {
...
 		FromIDExclusive string
 		ToIDInclusive   string
-		Limit           int
 		ExpectIDs       []string
 	}{
 		{
+			Desc:            "has no gaps - should omit nothing",
 			FromIDExclusive: "start",
 			ToIDInclusive:   "C",
 			ExpectIDs:       []string{"C", "B", "A"},
```
kegsay, line 1178: I find it weird that we reverse the ordering here, but whatever.

kegsay (approved):
Let's give it a go.

state/event_table_test.go
DMRobertson: `X..=Y` includes X, but I want to exclude X here.

state/event_table.go
DMRobertson: This is goland. I made the mistake of teaching it how I liked SQL to be formatted. Let me unteach it.

DMRobertson: image
this is presumably the culprit

DMRobertson: 3e02510

state/event_table_test.go
DMRobertson: I was also surprised!

DMRobertson: AFAICS the only call site of SelectLatestEventsBetween is

https://github.com/matrix-org/sliding-sync/blob/d3ba1f1c30017a1dbaaa654bbae8ba2e9c12661b/state/storage.go#L674-L689

so it won't be too hard to fix this up in the future, if we want to.

state/event_table_test.go
DMRobertson: Hmm. I think rename these to e.g. `chunkX-eventY` , so that gaps only appear when we change chunks.

tests-integration/poller_test.go
```diff
@@ -535,3 +534,68 @@ func TestPollerExpiryEnsurePollingRaceDoesntWedge(t *testing.T) {
...
+				TimelineLimit: 10,
+			},
+		},
+	})
+
+	t.Log("The response's timeline should only include the event after the gap.")
+	m.MatchResponse(t, res, m.MatchRoomSubscription(roomID,
+		m.MatchRoomTimeline([]json.RawMessage{msgAfterGap}),
+	))
+}
```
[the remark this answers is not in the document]
DMRobertson: > Which is _good_.

Indeed---just checking that the test really was proof we'd done this properly.

DMRobertson:
Only EW failed, but it suceeded on the prior commit. The only change I made was in unit tests, so I assume this is some Cypressy flake.

PR #301, opened 2023-09-13
Better propagate request context

sandhose:
This properly propagates the go Context on down to all HTTP calls, which means that outgoing request have the OTLP trace context.
This also adds the Jaeger propagator to the list of OTEL propagators, so that Synapse properly gets the incoming trace context.
It also upgrades all the OTEL libraries


kegsay (approved):
LGTM

PR #302, opened 2023-09-14
bugfix: ensure that we always call EnsurePolling for expired tokens

kegsay:
Otherwise, we might cache that the token is expired and never let a new token start the poller again.

### DIFF
--- a/sync3/handler/ensure_polling.go
+++ b/sync3/handler/ensure_polling.go
@@ -16,8 +16,8 @@ import (
 type pendingInfo struct {
 	// done is set to true when the EnsurePolling request received a response.
 	done bool
-	// success is true when done is true and EnsurePolling succeeded; otherwise false.
-	success bool
+	// expired is true when the token is expired. Any 'done'ness should be ignored for expired tokens.
+	expired bool
 	// ch is a dummy channel which never receives any data. A call to
 	// EnsurePoller.OnInitialSyncComplete will close the channel (unblocking any
 	// EnsurePoller.EnsurePolling calls which are waiting on it) and then set the ch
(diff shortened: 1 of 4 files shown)


sync3/handler/ensure_polling.go
```diff
@@ -58,19 +58,31 @@ func NewEnsurePoller(notifier pubsub.Notifier, enablePrometheus bool) *EnsurePol
...
+// the caller's responsibility to call OnInitialSyncComplete when new events arrive. Returns whether
+// or not the token is expired
 func (p *EnsurePoller) EnsurePolling(ctx context.Context, pid sync2.PollerID, tokenHash string) bool {
 	ctx, region := internal.StartSpan(ctx, "EnsurePolling")
 	defer region.End()
 	p.mu.Lock()
-	// do we need to wait?
-	if p.pendingPolls[pid].done {
+	// do we need to wait? Expired devices ALWAYS need a fresh poll
+	expired := p.pendingPolls[pid].expired
```
DMRobertson, line 68: If pid is missing from the map then expired is false. Is this correct? I think so?

DMRobertson: Oh it looks like this function isn't writing into the pendingPolls map at all(?)

sync3/handler/ensure_polling.go
```diff
@@ -129,7 +141,7 @@ func (p *EnsurePoller) OnInitialSyncComplete(payload *pubsub.V2InitialSyncComple
 		log.Trace().Msg("OnInitialSyncComplete: we weren't waiting for this")
```
DMRobertson, line 141: Should we promote this to a warning?

tests-integration/poller_test.go
```diff
@@ -461,3 +463,75 @@ func TestPollerExpiryEnsurePollingRace(t *testing.T) {
...
+
+// Regression test for the bugfix for https://github.com/matrix-org/sliding-sync/issues/287#issuecomment-1706522718
+// Specifically, we could cache the failure and never tell the poller about new tokens, wedging the client(!). This
+// seems to have been due to the following:
+//   - client hits sync for the first time. We /whoami and remember the token->user mapping in TokensTable.
+//   - client syncing + poller syncing, everything happy.
+//   - token expires. OnExpiredToken is sent to EnsurePoller which removes the entry from EnsurePoller and nukes the conns.
+//   - client hits sync, gets 400 M_UNKNOWN_POS due to nuked conns.
+//   - client hits a fresh /sync: for whatever reason, the token is NOT 401d there and then by the /whoami lookup failing.
+//     Maybe failed to remove the token, but don't see any logs to suggest this. Seems to be an OIDC thing.
```
DMRobertson, line 475: sad that we don't have a better understanding of this. Is there any logging we can add to make things less unclear for our future selves?

tests-integration/poller_test.go
```diff
@@ -461,3 +463,75 @@ func TestPollerExpiryEnsurePollingRace(t *testing.T) {
...
+	defer v3.close()
+
+	v2.addAccount(t, alice, aliceToken)
+
+	// Arrange the following:
+	// 1. A request arrives from an unknown token.
+	// 2. The API makes a /whoami lookup for the new token. That returns without error.
+	// 3. The old token expires.
+	// 4. The poller tries to call /sync but finds that the token has expired.
+	// NEW 5. Using a "new token" works.
```
DMRobertson, line 506: Does this make the previous test I wrote above redundant? I.e. is is the same stuff + new, or is there some difference in the initial setup/assertions too?

DMRobertson (approved):
:shipit:

DMRobertson:
Kegan thinks that https://github.com/matrix-org/sliding-sync/pull/302#discussion_r1326194984 needs thought, but he's going to the pub.

DMRobertson (changes requested):
see above

sync3/handler/ensure_polling.go
kegsay: Yes this spams on startup, so trace is fine.

tests-integration/poller_test.go
kegsay: nothing comes to mind immediately sadly.

tests-integration/poller_test.go
kegsay: It's the same stuff + NEW, which I copy/pasta'd. Having 2 is still useful insofar that if your one passes but this new one fails, we know it's due to the extra step.

sync3/handler/ensure_polling.go
kegsay: If pid is missing, then expired is false, and this is correct insofar that the if statement below should just care about if it is "done" or not, which will be false for missing pids. So yes, happy this is fine.

Issue #241, opened 2023-08-08
Room Heroes


pixlwave:
In order to render [stacked avatar clusters](https://www.figma.com/file/iIjwDl9HjaaF1vlDU96h8K/INBOX-%C2%B7-Compound-iOS%2FAndroid-Components?type=design&node-id=468-26319&mode=design&t=MXbTUt6Mu5rSViE7-4), it is desirable to have a flag/extension to request the Room's heroes be included in sync responses.

The clusters have the requirement that these will be displayed in a consistent order and it seems like we should use the hero ordering for this so it matches the generated room name. The linked designs scale up to 4 room members (including the current user), so in theory we only need the first 3 heroes for Element X (which might be a useful data saver instead of all 5)

pixlwave:
Adding a note that this will also be useful for 1:1 DM colour computation.

S7evinK:
(copy and pasting, with some modification)
- If we add a `heroes` flag, this would be per list and result in potentially wasted bandwidth (think of rooms where we have a `m.room.name` or already have an avatar URL)
- Should we try to be "intelligent" and automatically return `heroes` if the room name was calculated using heroes?
- Combination of the above? (as in: we add a `heroes` flag and only return `heroes` if we calculated the name using heroes)

imo, the last option sounds the way to go, as this doesn't return unused (by the client) data

poljar:
How about something like this for an explicitly named room:

```json
{
  "explicit": {
    "name": "my_name"
  }
}
```

And this for a calculated room:
```json
{
  "calculated": {
    "heroes": [
      "@foo:bar"
    ]
  }
}
```

Which would map to this Rust struct:

```rust
#[derive(Serialize, Deserialize)]
enum RoomName {
    #[serde(rename = "calculated")]
    Calculated { heroes: Vec<String> },
    #[serde(rename = "explicit")]
    Explicit { name: String },
}
```

We could use a [internally tagged] representation if you prefer that one, which seems to be more common in the Matrix world.

We could also go a bit further and try to map what we already have in the Rust SDK:
https://github.com/matrix-org/matrix-rust-sdk/blob/dde2f408c5e9449db3f01931ddae3cfd434c2d3d/crates/matrix-sdk-base/src/rooms/mod.rs#L32-L48

Though this enum has the problem that the heroes are flattened out into a string, so we would need to fix that.

I hope that this would also allow @pixlwave to do whatever he needs to do on the app side.

[...]

PR #304, opened 2023-09-15
Add posibility to request room heroes

S7evinK:
Fixes https://github.com/matrix-org/sliding-sync/issues/241

Draft until I update the MSC

This adds a `heroes` flag to `RoomSubscription`s. If the room name was calculated using room heroes, the response for the room also contains an array of `heroes`, containing the `user_id`, and optionally the fields `displayname` and `avatar_url`.

### DIFF
--- a/sync3/handler/connstate_live.go
+++ b/sync3/handler/connstate_live.go
@@ -262,7 +262,8 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 				roomName, calculated := internal.CalculateRoomName(metadata, 5) // TODO: customisable?
 
 				thisRoom.Name = roomName
-				if calculated {
+
+				if s.roomSubscriptions[roomUpdate.RoomID()].IncludeHeroes() && calculated {
 					thisRoom.Heroes = metadata.Heroes
 				}
 			}
--- a/sync3/handler/connstate_live.go
+++ b/sync3/handler/connstate_live.go
@@ -259,7 +259,12 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 			if delta.RoomNameChanged {
 				metadata := roomUpdate.GlobalRoomMetadata()
 				metadata.RemoveHero(s.userID)
-				thisRoom.Name = internal.CalculateRoomName(metadata, 5) // TODO: customisable?
+				roomName, calculated := internal.CalculateRoomName(metadata, 5) // TODO: customisable?
+
+				thisRoom.Name = roomName
+				if calculated {
+					thisRoom.Heroes = metadata.Heroes
+				}
 			}
 			if delta.RoomAvatarChanged {
 				metadata := roomUpdate.GlobalRoomMetadata()
--- a/tests-e2e/membership_transitions_test.go
+++ b/tests-e2e/membership_transitions_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/matrix-org/sliding-sync/sync3"
 	"github.com/matrix-org/sliding-sync/testutils/m"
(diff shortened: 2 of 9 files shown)


sync3/handler/connstate_live.go
```diff
@@ -259,7 +259,13 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 			if delta.RoomNameChanged {
 				metadata := roomUpdate.GlobalRoomMetadata()
 				metadata.RemoveHero(s.userID)
-				thisRoom.Name = internal.CalculateRoomName(metadata, 5) // TODO: customisable?
+				roomName, calculated := internal.CalculateRoomName(metadata, 5) // TODO: customisable?
+
+				thisRoom.Name = roomName
+
+				if s.roomSubscriptions[roomUpdate.RoomID()].IncludeHeroes() && calculated {
```
kegsay, line 266: This isn't enough as it won't work for lists. We need to instead find out which subscriptions this room relates to and use the values from that. Call `roomIDsToLists := s.lists.ListsByVisibleRoomIDs(s.muxedReq.Lists)` and see if any of those list keys have heroes included in `s.muxedReq`.

sync3/handler/connstate_live.go
```diff
@@ -444,3 +444,20 @@ func (s *connStateLive) resort(
 	}
 	return ops, hasUpdates
 }
+
+// shouldIncludeHeroes returns whether the given roomID is in a list or direct
+// subscription which should return heroes.
+func (s *connStateLive) shouldIncludeHeroes(roomID string) bool {
+	roomIDsToLists := s.lists.ListsByVisibleRoomIDs(s.muxedReq.Lists)
+	subscriptionInclude := s.roomSubscriptions[roomID].IncludeHeroes()
```
kegsay, line 452: Rejig this so it's faster. Checking the lists is expensive.
```go
if s.roomSubscriptions[roomID].IncludeHeroes() {
    return true
}
// rest of list checks
```

sync3/handler/connstate_live.go
```diff
@@ -263,7 +263,7 @@ func (s *connStateLive) processLiveUpdate(ctx context.Context, up caches.Update,
 
 				thisRoom.Name = roomName
 
-				if s.roomSubscriptions[roomUpdate.RoomID()].IncludeHeroes() && calculated {
+				if calculated && s.shouldIncludeHeroes(roomUpdate.RoomID()) {
```
kegsay, line 266: Looks better, but where's the tests?! :D

sync3/handler/connstate_live.go
S7evinK: Right, that's easier.

tests-e2e/membership_transitions_test.go
```diff
@@ -659,6 +660,47 @@ func TestHeroesOnMembershipChanges(t *testing.T) {
...
+						TimelineLimit: 10,
+					},
+				},
+			},
+		}, func(response *sync3.Response) error {
+			r, ok := response.Rooms[listRoomID]
+			if !ok {
+				return fmt.Errorf("room %q not in response", listRoomID)
+			}
+			// wait for bob to be joined
```
kegsay, line 683: Can't you just use `res := alice.SlidingSyncUntilMembership(t, "", listRoomID, bob, "join")`?

kegsay (approved):

tests-e2e/membership_transitions_test.go
S7evinK: For the previous test yes, but this test should use a list instead of a room subscription (used in `SlidingSyncUntilMembership`)

Issue #305, opened 2023-09-15
Mark expired tokens in the DB rather than DELETEing them


kegsay:
This is a performance optimisation.

In OIDC, when a token is expired, we delete the token from the tokens table. When the client hits the proxy with the same token later on, we then fail to find the token in the table and do a /whoami lookup which of course 401s. This is a wasted round trip and spams logs and makes things look broken when they aren't. We should instead:
 - Add an expired bool column to syncv3_sync2_tokens
 - Mark the token as expired rather than deleting the token.
 - On SELECT, if expired => send 401.
 - Periodicially clean the tokens table based on last_seen.

PR #307, opened 2023-09-18
Mark tokens as expired instead of deleting them 

S7evinK:
Fixes https://github.com/matrix-org/sliding-sync/issues/305

This PR changes the following:
- adds a `expired` column which defaults to `false`
- Sets the `expired` flag when we receive a 401 from upstream
- when setting up a connection and we know about the hash, but it is expired, returns 401 to the client
- starts a timer to remove expired tokens


sync2/tokens_table.go
```diff
@@ -42,17 +44,32 @@ func NewTokensTable(db *sqlx.DB, secret string) *TokensTable {
...
 
-	return &TokensTable{
+	t := &TokensTable{
 		db:     db,
 		key256: hash.Sum(nil),
 	}
+
+	var deleteExpiredTokens func()
+	deleteExpiredTokens = func() {
+		deleted, err := t.deleteExpiredTokensAfter(30 * 24 * time.Hour)
```
S7evinK, line 62: 30 days might be a bit much for OIDC, so maybe a bit lower?

DMRobertson: FWIW I used 30 days in https://github.com/matrix-org/sliding-sync/issues/242 for a different (but related) purpose.

sync2/tokens_table.go
```diff
@@ -42,17 +44,32 @@ func NewTokensTable(db *sqlx.DB, secret string) *TokensTable {
...
+		if err != nil {
+			logger.Warn().Err(err).Msg("failed to delete expired tokens")
+			return
+		}
+		logger.Trace().Int64("deleted", deleted).Msg("deleted expired tokens")
+		time.AfterFunc(time.Hour, deleteExpiredTokens)
+	}
+	time.AfterFunc(time.Hour, deleteExpiredTokens)
```
kegsay, line 70: Please don't tie this to the `New...` constructor. This is going to be such a faff to work with in tests otherwise.

sync2/tokens_table.go
```diff
@@ -42,17 +44,32 @@ func NewTokensTable(db *sqlx.DB, secret string) *TokensTable {
...
 		db:     db,
 		key256: hash.Sum(nil),
 	}
+
+	var deleteExpiredTokens func()
+	deleteExpiredTokens = func() {
+		deleted, err := t.deleteExpiredTokensAfter(30 * 24 * time.Hour)
+		if err != nil {
+			logger.Warn().Err(err).Msg("failed to delete expired tokens")
+			return
```
kegsay, line 65: Doesn't make sense to then never delete ever again :S surely we should retry in 1h?

sync2/tokens_table.go
```diff
@@ -258,3 +275,38 @@ func (t *TokensTable) Delete(accessTokenHash string) error {
 	}
```
kegsay, line 275: `func (t *TokensTable) Delete(accessTokenHash string) error {` is unused now then? Remove it please.

sync3/handler/handler.go
```diff
@@ -360,6 +360,16 @@ func (h *SyncLiveHandler) setupConnection(req *http.Request, syncReq *sync3.Requ
 			}
 		}
 	}
+	// We know about this token, but is expired, let the client know to refresh their token.
+	if token.Expired {
+		return req, nil, &internal.HandlerError{
+			StatusCode: http.StatusUnauthorized,
+			ErrCode:    "M_UNKNOWN_TOKEN",
+			// not exactly /whoami, but would be if we didn't keep the expired state
+			Err: fmt.Errorf("/whoami returned HTTP 401"),
```
kegsay, line 369: Please don't lie else this will be confusing when we see this in bug reports. Say "token marked as expired" instead.

sync2/tokens_table_test.go
```diff
@@ -128,7 +129,61 @@ func TestDeletingTokens(t *testing.T) {
...
+	var token *Token
+	err := sqlutil.WithTransaction(db, func(txn *sqlx.Tx) (err error) {
+		token, err = tokens.Insert(txn, accessToken, "@bob:builders.com", "device", time.Now())
+		if err != nil {
+			t.Fatalf("Failed to Insert token: %s", err)
+		}
+		return nil
+	})
+	t.Log("We should be able to fetch this token without error.")
+	_, err = tokens.Token(accessToken)
```
kegsay, line 153: Verify the token is not expired.

sync2/tokens_table_test.go
```diff
@@ -128,7 +129,61 @@ func TestDeletingTokens(t *testing.T) {
...
+	}
+
+	t.Log("Does not return an error if the hash can not be found")
+	err = tokens.Expire("idontexist")
+	if err != nil {
+		t.Fatalf("Expected no error for non-existent hash, got %s", err)
+	}
+
+	t.Log("Deletes expired tokens")
+	deleted, err := tokens.deleteExpiredTokensAfter(time.Nanosecond)
```
kegsay, line 181: I'd check before this that `tokens.deleteExpiredTokensAfter(time.Hour)` has `deleted == 0` i.e it didn't delete anything.

kegsay (approved):
Needs some tweaks but otherwise this is great.

kegsay:
> With the [failing test](https://github.com/matrix-org/sliding-sync/actions/runs/6223030659/job/16888107341?pr=307#step:7:33), I'm wondering the same as @DMRobertson here [#302 (comment)](https://github.com/matrix-org/sliding-sync/pull/302#discussion_r1326224448) The new test added in #302 is happy here, but the "old" one is failing.

This isn't quite right. The failing test is `TestPollersCanBeResumedAfterExpiry`. The tests referred to in #302 are `TestPollerExpiryEnsurePollingRace` and `TestPollerExpiryEnsurePollingRaceDoesntWedge`.

The failing test is 401ing alice's token because it is old as per comments:
```go
// Inject an old token from Alice and a new token from Bob into the DB.
```
It then calls `h.callbacks.OnExpiredToken(context.Background(), hashToken(p.accessToken), p.userID, p.deviceID)` due to the test calling:
```
	t.Log("Manually trigger a poller cleanup.")
	v3.h2.ExpireOldPollers()
```
So the token for alice will be marked as expired in the DB in this PR, hence the 401. The failing section of test is:
```go
	t.Log("Alice makes a new sliding sync request")
	res := v3.mustDoV3Request(t, aliceToken, sync3.Request{
		Extensions: extensions.Request{
			AccountData: &extensions.AccountDataRequest{
				Core: extensions.Core{
					Enabled: &boolTrue,
				},
			},
		},
	})

	t.Log("Alice's poller should have been polled.")
	v2.waitUntilEmpty(t, aliceToken)
```
Which makes sense as the expired `aliceToken` won't hit `/whoami` or anything anymore, it'll just immediately 401. The test needs to be updated to use a new token for Alice.

S7evinK:
@kegsay is it still worth taking a look at this or can we close this?

kegsay:
I think we can close this for now. We may want to revisit this later on.

PR #311, opened 2023-09-21
Reduce mutex contention on callbacks

kegsay:
Don't keep `pollerMu` held for longer than we need to.


DMRobertson:
I'd like to read through the rest of the places that acquire that mutex---I bet there are other places we can do better.

DMRobertson (approved):
Happy if CI is.

PR #323, opened 2023-10-02
Add issue templates

DMRobertson:
These are mostly the standard github templates with a few extra prods from me.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,31 @@
+---
+name: Bug
+about: Something in the proxy is misbehaving
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behaviour:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behaviour**
+A clear and concise description of what you expected to happen.
+
+**Relevant logs**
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,31 @@
...
+about: Something in the proxy is misbehaving
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
```
DMRobertson, line 12: Todo: get the reporter to tell us their homeserver address, or at least if they're self-hosted.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,31 @@
...
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behaviour**
+A clear and concise description of what you expected to happen.
+
+**Relevant logs**
+Please include any relevant logs from the sliding sync proxy. Enclose them in triple backquotes, e.g.
```
DMRobertson, line 24: Todo: ask them if they have sent a rageshake from Element X.

.github/ISSUE_TEMPLATE/bug.md
DMRobertson: Todo: tell them to open an issue on the client repo if they think this is a problem with EX rather than the proxy.

DMRobertson: Todo: get them to say what proxy version they're running.

kegsay:
In practice we don't get flooded with badly formed issues, so this feels unnecessary in its current form. Will revisit if this changes.

PR #324, opened 2023-10-02
Remove `RequestedLatestEvents` field from `UserRoomData`

DMRobertson:
Helps with the sort of confusion I had complained about in https://github.com/matrix-org/sliding-sync/issues/46.

My goal here is to make the UserRoomData type contain exactly the data that the UserCache is responsible for maintaining. This is step 1 of N, for some integer N > 1.

### DIFF
--- a/sync3/caches/user.go
+++ b/sync3/caches/user.go
@@ -347,6 +347,21 @@ func (c *UserCache) LoadRoomData(roomID string) UserRoomData {
 	return data
 }
 
+// LoadRooms is a batch version of LoadRoomData. Returns a map keyed by roomID.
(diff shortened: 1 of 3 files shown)


kegsay (approved):
LGTM. I had thought maybe the UserRoomData/timelines were got atomically with the same lock but no, it doesn't. Timelines are pinned on the loadPos and URDs are just whatever is in the user cache at the time.

DMRobertson:
> I had thought maybe the UserRoomData/timelines were got atomically with the same lock but no, it doesn't. Timelines are pinned on the loadPos and URDs are just whatever is in the user cache at the time.

I had the same thought, but concluded that this was no worse than what we have today. Note that the global metadata is also just whatever happens to be in-memory too.

PR #333, opened 2023-10-10
Cleanup invites in the poller, not the API

DMRobertson:
Kegan and I were surprised to see the API side of the proxy doing this.

I removed the existing logic and saw that TestStuckInvites broke. Hopefully it should pass after these changes. Does that count as sufficient testing?

### DIFF
--- a/state/invites_table.go
+++ b/state/invites_table.go
@@ -3,6 +3,7 @@ package state
 import (
 	"database/sql"
 	"encoding/json"
+	"github.com/lib/pq"
 
 	"github.com/jmoiron/sqlx"
 )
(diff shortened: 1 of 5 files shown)


state/invites_table.go
```diff
@@ -47,6 +51,40 @@ func (t *InvitesTable) RemoveInvite(userID, roomID string) error {
...
+	}
+
+	var usersToRemove []string
+	for userID, membership := range memberships {
+		if membership != "invite" && membership != "_invite" {
+			usersToRemove = append(usersToRemove, userID)
+		}
+	}
+
+	_, err := txn.Exec(`
```
kegsay, line 80: `if len(usersToRemove) == 0 { return nil }`

kegsay (approved):

state/invites_table.go
DMRobertson: Ahh yes, thanks.

PR #337, opened 2023-10-11
Fix race condition in test

kegsay:
Implements half of #243 in a different way which involves less error prone locking.

### DIFF
--- a/sync2/device_data_ticker_test.go
+++ b/sync2/device_data_ticker_test.go
@@ -10,12 +10,31 @@ import (
 	"github.com/matrix-org/sliding-sync/pubsub"
 )
 
+type syncSlice[T any] struct {
+	slice []T
+	mu    sync.Mutex
+}
+
+func (s *syncSlice[T]) append(item T) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.slice = append(s.slice, item)
+}
+
+func (s *syncSlice[T]) clone() []T {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	result := make([]T, len(s.slice))
+	copy(result, s.slice)
+	return result
+}
+
 func TestDeviceTickerBasic(t *testing.T) {
 	duration := time.Millisecond
 	ticker := NewDeviceDataTicker(duration)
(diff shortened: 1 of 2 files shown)


sync2/device_data_ticker_test.go
```diff
@@ -10,12 +10,31 @@ import (
...
+	s.slice = append(s.slice, item)
+}
+
+func (s *syncSlice[T]) clone() []T {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	result := make([]T, len(s.slice))
+	copy(result, s.slice)
+	return result
+}
```
DMRobertson, line 30: is there no sync.Slice the same way there is sync.Map?! odd

sync2/device_data_ticker_test.go
```diff
@@ -31,29 +50,31 @@ func TestDeviceTickerBasic(t *testing.T) {
 		DeviceID: "b",
 	})
 	time.Sleep(duration * 2)
-	if len(payloads) != 1 {
-		t.Fatalf("expected 1 callback, got %d", len(payloads))
+	result := payloads.clone()
```
DMRobertson, line 53: Why do we need to clone here, sorry?

DMRobertson (approved):

sync2/device_data_ticker_test.go
kegsay: We'd need to lock every time we access the underlying slice, which could then lead to inconsistent results (e.g len=1 when len() is called, but then increases to len=2 later on. It's easier to snapshot the slice.

sync2/poller_test.go
```diff
@@ -18,6 +18,31 @@ import (
...
+var (
+	timeSinceMu    sync.Mutex
+	timeSinceValue = time.Duration(0) // 0 means use the real impl
+)
+
+func setTimeSinceValue(val time.Duration) {
+	timeSinceMu.Lock()
+	timeSinceValue = time.Minute * 2
+	timeSinceMu.Unlock()
+}
```
DMRobertson, line 34: val seems unused here?

PR #339, opened 2023-10-11
Build and attach {amd,arm}64 binaries to releases

### LINKED ISSUE #334: Release build artifacts on releases
This proxy is required for upcoming ElementX. Make it easier for the community to use the proxy in the future:

- Release go build artifacts when tagging releases

It seems possible: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go

DMRobertson:
Completely untested. Cribbed a little [from Synapse's CI](https://github.com/matrix-org/synapse/blob/139a24de9ee0e81faece1e375a197123a6e10b67/.github/workflows/release-artifacts.yml#L186-L212) but kept simple.

I suggest 
1. we commit this to main
2. I create a draft release pointing at main
3. We check the CI runs
4. If anything breaks, discard the release, fix and return to (1)
5. ???
6. Profit

I also ran this through https://github.com/rhysd/actionlint which reported no complaints.

Fixes #334.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        os: ["linux"]
         arch: ["amd64", "arm64"]
     steps:
       - uses: actions/checkout@v4
@@ -18,6 +19,7 @@ jobs:
       - run: mkdir build
       - run: go build -o build/syncv3_linux_${{ matrix.arch }} ./cmd/syncv3
         env:
+          GOOS: ${{ matrix.os }}
           GOARCH: ${{ matrix.arch }}
       - uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v1
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: "Build and attach binary to release"
+on:
+  release:
+    types: ["created"]
+permissions:
+  contents: write  # to upload the binaries to the release
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["amd64", "arm64"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.20"
+      - run: mkdir build
+      - run: go build ./cmd/syncv3 -o build/syncv3_linux_${{ matrix.arch }}
+        env:
+          GOARCH: ${{ matrix.arch }}
+      - uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v1
+        with:
+          files: build/syncv3_linux_${{ matrix.arch }}
+          fail_on_unmatched_files: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.20"
+      - run: mkdir build
+      - run: go build ./cmd/syncv3 -o build/syncv3_linux_${{ matrix.arch }}
+        env:
+          GOARCH: ${{ matrix.arch }}
+      - uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v1
```
DMRobertson, line 22: Synapse uses this. There is the official-looking https://github.com/actions/upload-release-asset, but this has been archived and now recomments softprops' action.

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+  release:
+    types: ["created"]
+permissions:
+  contents: write  # to upload the binaries to the release
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["amd64", "arm64"]
```
DMRobertson, line 12: I'm assuming it's quick enough to compile for arm. IME people are going to request arm binaries if we publish any kind of binary so we may as well do this now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
+name: "Build and attach binary to release"
+on:
+  release:
+    types: ["created"]
```
DMRobertson, line 4: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

FTR the types are

- published
- unpublished
- created
- edited
- deleted
- prereleased
- released

.github/workflows/release.yml
kegsay: No x86_64?

DMRobertson: amd64 and x86_64 are the same thing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+        arch: ["amd64", "arm64"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.20"
+      - run: mkdir build
+      - run: go build -o build/syncv3_linux_${{ matrix.arch }} ./cmd/syncv3
+        env:
+          GOARCH: ${{ matrix.arch }}
```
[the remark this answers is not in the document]
DMRobertson: I assumed that would be inherited from the fact that we're running on ubuntu, but I can make that explicit.

DMRobertson:
Oh ffs:

> Note: Workflows are not triggered for the created, edited, or deleted activity types for draft releases. When you create your release through the GitHub browser UI, your release may automatically be saved as a draft.

from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

DMRobertson:
Which contradicts https://docs.github.com/en/webhooks/webhook-events-and-payloads#release

> Action type: created
> A draft was saved, or a release or pre-release was published without previously being saved as a draft.

DMRobertson:
https://github.com/orgs/community/discussions/7118

PR #340, opened 2023-10-11
Use Complement libraries

kegsay:
An experiement in using bits of Complement. Doesn't use all the machinery, as we don't deploy new containers for each test, but it's a start (uses must/match/client).


DMRobertson:
> Doesn't use all the machinery, as we don't deploy new containers for each test, but it's a start (uses must/match/client).

Do we even want to do this? While we do have tests that leak state, the speed of running them locally is really valuable and it would be sad to have to wait for Synapse's docker container to start up over and over again.

tests-e2e/client_test.go
```diff
@@ -38,635 +26,47 @@ type Event struct {
...
-// This function should not be used to set an "access_token" parameter for Matrix authentication.
-// Instead, set CSAPI.AccessToken.
-func WithQueries(q url.Values) RequestOpt {
-	return func(req *http.Request) {
-		req.URL.RawQuery = q.Encode()
-	}
+	*client.CSAPI
+	Localpart string
+	Domain    string
+	AvatarURL string
```
DMRobertson, line 43: Should this be folded into complement?

DMRobertson (approved):

tests-e2e/client_test.go
kegsay: Maybe. Unsure for now, so leaving it here.

kegsay:
> Do we even want to do this? 

Maybe in the future when we reuse containers. Right now, no.

PR #329, opened 2023-10-04
Second attempt to update the u

DMRobertson:
https://github.com/matrix-org/sliding-sync/pull/310 was getting out of control. I did an interactive rebase, threw some things away, and wrote some new things. This is the result.

This PR:

1. Removes the mechanism of prepending state events to the timeline, creating new snapshots instead.
2. Emits cache invalidation messages when doing so.
3. Updates the user cache in response to an invalidation payload.

(3) was the hard part. We could maybe land it separately, but we need (3) before (1) and (2).

PR #347, opened 2023-10-19
Fix keys() return value starting with zero-valued keys

DMRobertson:
Also pull this into `internal` so that other packages can reuse it

Hopefully we can use
https://pkg.go.dev/golang.org/x/exp@v0.0.0-20231006140011-7918f672742d/maps#Keys one day

Pulled out of #329.

### DIFF
--- a/internal/util_test.go
+++ b/internal/util_test.go
@@ -25,6 +25,6 @@ func assertSlice(t *testing.T, got, want []string) {
 	sort.Slice(want, func(i, j int) bool { return want[i] < want[j] })
 
 	if !reflect.DeepEqual(got, want) {
-		t.Errorf("After sorting, got %v ubt expected %v", got, want)
+		t.Errorf("After sorting, got %v but expected %v", got, want)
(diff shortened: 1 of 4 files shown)


internal/util_test.go
```diff
@@ -0,0 +1,30 @@
...
+func assertSlice(t *testing.T, got, want []string) {
+	if len(got) != len(want) {
+		t.Errorf("got length %d, expected length %d", len(got), len(want))
+	}
+
+	sort.Slice(got, func(i, j int) bool { return got[i] < got[j] })
+	sort.Slice(want, func(i, j int) bool { return want[i] < want[j] })
+
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("After sorting, got %v ubt expected %v", got, want)
```
kegsay, line 28:
```suggestion
		t.Errorf("After sorting, got %v but expected %v", got, want)
```

kegsay (approved):

PR #351, opened 2023-10-19
Add migration to cleanup dead invites

### LINKED ISSUE #350: Invites are not always correctly handled over federation
When a user joins a room they were invited to _over federation_, the join event can lack an `unsigned.prev_content.membership == "invite"`. Instead, the `prev_content` is missing entirely. @DMRobertson and I theorise this is because the receiving server could not get the `replaces_state` event, because that would require a backfill.

Unfortunately, up until https://github.com/matrix-org/sliding-sync/pull/333 the proxy was relying on this field being set. If it was unset, the invite would never be removed from the invites table.

When a client who is joined-but-still-has-the-invite requests information about this room, they end up with a blend of joined/invited data. Specifically, the `invite_state` is set to the stripped invite state events from the invite. This is bad because the presence of `invite_state` tells clients that the user is _invited_ and not joined to the room. This could cause them to insert the room into a different place in the UI, or remove it entirely from the room list.

The fix in #333 means we no longer rely on this field, but there are still rooms which may be in this state in the database. We need to write a migration which:
 - For each invite in the invites table (which shouldn't be too big relative to the number of joined rooms)
    * Is this user present in the current state of the room? NO: ignore (the proxy isn't in this room yet) YES:
    * Is the membership value `invite`? YES: ignore (someone else joined the room and the membership value is from their poller). NO:
    * Remove the invite from the invites table, as their current membership is something other than `invite`.

DMRobertson:
Fixes #350.

### DIFF
--- a/state/migrations/20231019153023_cleanup_dead_invites_test.go
+++ b/state/migrations/20231019153023_cleanup_dead_invites_test.go
@@ -97,7 +97,7 @@ func TestDeadInviteCleanup(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	assertInvites(t, "alice,", aliceInvites, map[string][]json.RawMessage{
+	assertInvites(t, "alice", aliceInvites, map[string][]json.RawMessage{
 		"!y": emptyInviteState,
 	})
 
@@ -105,7 +105,7 @@ func TestDeadInviteCleanup(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	assertInvites(t, "bob,", bobInvites, map[string][]json.RawMessage{
+	assertInvites(t, "bob", bobInvites, map[string][]json.RawMessage{
 		"!x": emptyInviteState,
 	})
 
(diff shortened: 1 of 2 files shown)


kegsay (approved):
Looks good when SELECTing on m.org - needs tests though.

state/migrations/20231019153023_cleanup_dead_invites_test.go
```diff
@@ -0,0 +1,127 @@
...
+var deadInviteMigration embed.FS
+
+func TestDeadInviteCleanup(t *testing.T) {
+	db, close := connectToDB(t)
+	defer close()
+	store := state.NewStorageWithDB(db, false)
+
+	// Alice and Bob have invites to rooms X and Y.
+	// Chris has an invite to room Z.
+	t.Log("Alice is joined to room X; Bob to room Y. Chris isn't joined to any rooms.")
```
kegsay, line 24: Log line is confusing and wrong? Wrong position in the test and Bob is banned from room Y, not joined.

kegsay (approved):

state/migrations/20231019153023_cleanup_dead_invites_test.go
DMRobertson: Oops. Thanks.

PR #355, opened 2023-10-26
Cancel outstanding requests when destroying conns

DMRobertson:
Will help with #294.

### DIFF
--- a/sync3/handler/connstate.go
+++ b/sync3/handler/connstate.go
@@ -23,8 +23,9 @@ type ConnState struct {
 	userID   string
 	deviceID string
 	// the only thing that can touch these data structures is the conn goroutine
-	muxedReq *sync3.Request
-	lists    *sync3.InternalRequestLists
+	muxedReq        *sync3.Request
+	cancelLatestReq context.CancelFunc
+	lists           *sync3.InternalRequestLists
 
 	// Confirmed room subscriptions. Entries in this list have been checked for things like
 	// "is the user joined to this room?" whereas subscriptions in muxedReq are untrusted.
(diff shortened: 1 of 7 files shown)


tests-e2e/handler_test.go
```diff
@@ -0,0 +1,85 @@
...
+	response := <-done
+
+	// TODO: At first I expected that this the cancelled request should return 400 M_UNKNOWN_POS.
+	// But I think that is best handled on the next incoming request.
+	// assertEqual(t, "status code", response.res.StatusCode, http.StatusBadRequest)
+	// assertEqual(t, "response errcode", response.body.Get("errcode").Str, "M_UNKNOWN_POS")
+
+	if response.duration > cancelWithin {
+		t.Errorf("Waited for %s, but expected second sync to cancel after at most %s", response.duration, cancelWithin)
+	}
```
DMRobertson, line 84: This was my attempt to detect if the request was cancelled early.

I did wonder about explicitly making a request with client-created context. But I wasn't sure if http.Client would mark the context as Done when the request ends.

DMRobertson: > But I wasn't sure if http.Client would mark the context as Done when the request ends.

seemingly not. That makes the test much cleaner.

sync3/handler/connstate.go
```diff
@@ -723,6 +724,8 @@ func (s *ConnState) trackProcessDuration(ctx context.Context, dur time.Duration,
 // Called when the connection is torn down
 func (s *ConnState) Destroy() {
 	s.userCache.Unsubscribe(s.userCacheID)
+	logger.Debug().Str("user_id", s.userID).Str("device_id", s.deviceID).Msg("cancelling any in-flight requests")
+	s.cancelLatestReq()
```
kegsay, line 728: Nil check?

kegsay (approved):

PR #358, opened 2023-10-28
add (docker-)compose example

violoncelloCH:
This adds a compose.yml sample config to easily deploy a slidingsync proxy accompanied by a postgresql db container.

### Pull Request Checklist

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -5,7 +5,7 @@ networks:
 
 services:
   slidingsync:
-    image: ghcr.io/matrix-org/sliding-sync
+    image: ghcr.io/matrix-org/sliding-sync:v0.99.13
     container_name: slidingsync
     environment:
       - SYNCV3_SERVER=https://matrix-client.matrix.org
@@ -13,7 +13,8 @@ services:
       - SYNCV3_BINDADDR=:8009
       - SYNCV3_DB=user=syncv3 dbname=syncv3 sslmode=disable host=slidingsync-postgres password=sliding-sync-v3
     depends_on:
-      - slidingsync-postgres
+      slidingsync-postgres:
+        condition: service_healthy
     networks:
       - sliding-sync
     ports:
@@ -24,8 +25,12 @@ services:
     image: "postgres:13"
     restart: always
     stop_signal: SIGINT
+    healthcheck:
+      test: ["CMD", "pg_isready"]
+      interval: 5s
+      timeout: 3s
+      retries: 5
     environment:
-      - POSTGRES_INITDB_ARGS=--data-checksums
       - POSTGRES_DB=syncv3
       - POSTGRES_USER=syncv3
       - POSTGRES_PASSWORD=sliding-sync-v3
--- a/compose.yml
+++ b/compose.yml
@@ -1,0 +1,35 @@
+version: "3.5"
+
+networks:
+  sliding-sync:
+
+services:
+  slidingsync:
+    image: ghcr.io/matrix-org/sliding-sync
+    container_name: slidingsync
+    environment:
+      - SYNCV3_SERVER=https://matrix-client.matrix.org
+      - SYNCV3_SECRET=  #use `openssl rand -hex 32` to generate a secret to paste here
+      - SYNCV3_BINDADDR=:8009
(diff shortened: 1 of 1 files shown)


compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    ports:
+      - "8009:8009"
+    restart: unless-stopped
+
+  slidingsync-postgres:
+    image: "postgres:13"
+    restart: always
+    stop_signal: SIGINT
+    environment:
+      - POSTGRES_INITDB_ARGS=--data-checksums
```
kegsay, line 28: Why?

compose.yml
```diff
@@ -0,0 +1,35 @@
+version: "3.5"
+
+networks:
+  sliding-sync:
+
+services:
+  slidingsync:
+    image: ghcr.io/matrix-org/sliding-sync
```
kegsay, line 8: Probably should have a versioned tag.

kegsay (changes requested):

compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    depends_on:
+      - slidingsync-postgres
+    networks:
+      - sliding-sync
+    ports:
+      - "8009:8009"
+    restart: unless-stopped
+
+  slidingsync-postgres:
+    image: "postgres:13"
```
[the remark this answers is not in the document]
violoncelloCH: ohh sure, I'll add that!

compose.yml
violoncelloCH: whoops, didn't notice - leftover from the template I copied... I'll remove it

compose.yml
violoncelloCH: this means that it has to be incremented regularly, but it makes sense obviously to have it... though I was unsure to how to handle this before a stable version of the project. As soon as there are major versions I guess we can simply pin it to a major version...
should I add an exact version tag for now ?

compose.yml
```diff
@@ -0,0 +1,35 @@
...
+networks:
+  sliding-sync:
+
+services:
+  slidingsync:
+    image: ghcr.io/matrix-org/sliding-sync
+    container_name: slidingsync
+    environment:
+      - SYNCV3_SERVER=https://matrix-client.matrix.org
+      - SYNCV3_SECRET=  #use `openssl rand -hex 32` to generate a secret to paste here
```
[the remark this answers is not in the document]
violoncelloCH: I'm actually using this in production, but I guess it can suit well for development too...
I'd leave it up to you to if you as project maintainer if you want to include a pre-generated secret for a dev-setup that would have to be changed for production use or leave the field empty.
for defaulting to matrix.org: I did this as the Readme defaults to matrix.org as well (docker command in Readme section - that's what I've used as a base to build this compose.yml)... I personally use this with my own homeserver to test-drive Element X... let me know if you want me to remove the default :)

compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      - SYNCV3_BINDADDR=:8009
+      - SYNCV3_DB=user=syncv3 dbname=syncv3 sslmode=disable host=slidingsync-postgres password=sliding-sync-v3
+    depends_on:
+      - slidingsync-postgres
```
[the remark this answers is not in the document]
violoncelloCH: thanks for the recommendation! I'll change this :)

violoncelloCH:
thanks for the review @kegsay ! and sorry for the slow response time - have been busy with university work the last weeks...

kegsay (commented):
I remain concerned that this is in a weird place where it doesn't quite target production deployments but also doesn't quite target testing deployments.

If we want to have this targeting production deployments then we need more parameterisation and some CI which runs this to ensure it doesn't break. 

If we want this to target testing deployments then some comments around things to change (version tag, secret, postgres password) should be all that's left.

I'm leaning towards targeting testing deployments so people can just `docker-compose up` and get a working sliding sync instance. This means we need:
 - to mention that you can `docker-compose up` in the README
 - add comments around the things which aren't safe to have in production.

kegsay:
Closing this due to inactivity. Please open another PR if you wish to target testing deployments.

I'm not opposed to having a docker-compose for simplicity, but we must be clear who the target audience is so we can make sure the defaults/exposed configuration are appropriate.

PR #363, opened 2023-11-03
Handle gappy state with a single snapshot, attempt 3

### LINKED ISSUE #232: Checkpoint room state

### LINKED ISSUE #294: Cache invalidation

DMRobertson:
A third iteration of "deal with the gappy poll problem".

I pulled out all the independent bits from #329 to other PRs. What was left fell into two categories: a) the original logic for trying to push out live updates with the gappy poll changes; and b) the logic for nuking connections. This PR consists of b) but not a).

Ancestry: #310 -> #329 -> this.
Closes #294.
Closes #232.

### DIFF
--- a/state/accumulator.go
+++ b/state/accumulator.go
@@ -141,69 +141,62 @@ type InitialiseResult struct {
 	// AddedEvents is true iff this call to Initialise added new state events to the DB.
 	AddedEvents bool
 	// SnapshotID is the ID of the snapshot which incorporates all added events.
-	// It has no meaning if AddedEvents is False.
+	// It has no meaning if AddedEvents is false.
 	SnapshotID int64
-	// PrependTimelineEvents is empty if the room was not initialised prior to this call.
-	// Otherwise, it is an order-preserving subset of the `state` argument to Initialise
-	// containing all events that were not persisted prior to the Initialise call. These
-	// should be prepended to the room timeline by the caller.
-	PrependTimelineEvents []json.RawMessage
+	// ReplacedExistingSnapshot is true when we created a new snapshot for the room and
+	// there a pre-existing room snapshot. It has no meaning if AddedEvents is false.
+	ReplacedExistingSnapshot bool
 }
 
-// Initialise starts a new sync accumulator for the given room using the given state as a baseline.
+// Initialise processes the state block of a V2 sync response for a particular room. If
+// the state of the room has changed, we persist any new state events and create a new
+// "snapshot" of its entire state.
 //
-// This will only take effect if this is the first time the v3 server has seen this room, and it wasn't
-// possible to get all events up to the create event (e.g Matrix HQ).
-// This function:
-// - Stores these events
-// - Sets up the current snapshot based on the state list given.
+// Summary of the logic:
 //
-// If the v3 server has seen this room before, this function
-//   - queries the DB to determine which state events are known to th server,
-//   - returns (via InitialiseResult.PrependTimelineEvents) a slice of unknown state events,
+//  0. Ensure the state block is not empty.
 //
-// and otherwise does nothing.
+//  1. Capture the current snapshot ID, possibly zero. If it is zero, ensure that the
+//     state block contains a `create event`.
+//
+//  2. Insert the events. If there are no newly inserted events, bail. If there are new
+//     events, then the state block has definitely changed. Note: we ignore cases where
+//     the state has only changed to a known subset of state events (i.e in the case of
+//     state resets, slow pollers) as it is impossible to then reconcile that state with
+//     any new events, as any "catchup" state will be ignored due to the events already
+//     existing.
+//
+//  3. Fetch the current state of the room, as a map from (type, state_key) to event.
+//     If there is no existing state snapshot, this map is the empty map.
+//     If the state hasn't altered, bail.
+//
+//  4. Create new snapshot. Update the map from (3) with the events in `state`.
+//     (There is similar logic for this in Accumulate.)
+//     Store the snapshot. Mark the room's current state as being this snapshot.
+//
+//  5. Any other processing of the new state events.
+//
+//  6. Return an "AddedEvents" bool (if true, emit an Initialise payload) and a
+//     "ReplacedSnapshot" bool (if true, emit a cache invalidation payload).
+
 func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (InitialiseResult, error) {
 	var res InitialiseResult
+	var startingSnapshotID int64
+
+	// 0. Ensure the state block is not empty.
 	if len(state) == 0 {
 		return res, nil
 	}
-	err := sqlutil.WithTransaction(a.db, func(txn *sqlx.Tx) error {
+	err := sqlutil.WithTransaction(a.db, func(txn *sqlx.Tx) (err error) {
+		// 1. Capture the current snapshot ID, checking for a create event if this is our first snapshot.
+
 		// Attempt to short-circuit. This has to be done inside a transaction to make sure
 		// we don't race with multiple calls to Initialise with the same room ID.
-		snapshotID, err := a.roomsTable.CurrentAfterSnapshotID(txn, roomID)
+		startingSnapshotID, err = a.roomsTable.CurrentAfterSnapshotID(txn, roomID)
 		if err != nil {
-			return fmt.Errorf("error fetching snapshot id for room %s: %s", roomID, err)
+			return fmt.Errorf("error fetching snapshot id for room %s: %w", roomID, err)
 		}
-		if snapshotID > 0 {
-			// Poller A has received a gappy sync v2 response with a state block, and
-			// we have seen this room before. If we knew for certain that there is some
-			// other active poller B in this room then we could safely skip this logic.
-
-			// Log at debug for now. If we find an unknown event, we'll return it so
-			// that the poller can log a warning.
-			logger.Debug().Str("room_id", roomID).Int64("snapshot_id", snapshotID).Msg("Accumulator.Initialise called with incremental state but current snapshot already exists.")
-			eventIDs := make([]string, len(state))
-			eventIDToRawEvent := make(map[string]json.RawMessage, len(state))
-			for i := range state {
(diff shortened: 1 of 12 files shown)


state/accumulator.go
```diff
@@ -214,77 +207,76 @@ func (a *Accumulator) Initialise(roomID string, state []json.RawMessage) (Initia
...
+		var currentState stateMap
+		if startingSnapshotID > 0 {
+			currentState, err = a.stateMapAtSnapshot(txn, startingSnapshotID)
+			if err != nil {
+				return fmt.Errorf("failed to load state map: %w", err)
+			}
+		} else {
+			currentState = stateMap{
+				Memberships: make(map[string]int64, len(events)),
+				Other:       make(map[[2]string]int64, len(events)),
```
kegsay, line 259: This will nearly always be way too large as membership events >>>> other events, I'd just leave memberships with `len(events)` and let `Other` self-size.

sync2/handler2/handler.go
```diff
@@ -370,20 +370,24 @@ func (h *Handler) Accumulate(ctx context.Context, userID, deviceID, roomID strin
...
 		internal.GetSentryHubFromContextOrDefault(ctx).CaptureException(err)
-		return nil, err
+		return err
 	}
-	if res.AddedEvents {
+	if res.ReplacedExistingSnapshot {
+		h.v2Pub.Notify(pubsub.ChanV2, &pubsub.V2InvalidateRoom{
+			RoomID: roomID,
+		})
+	} else if res.AddedEvents {
```
kegsay, line 384: Are they mutally exclusive? If so, an enum would be better than 2 bools.

sync3/handler/handler.go
```diff
@@ -818,7 +834,46 @@ func (h *SyncLiveHandler) OnInvalidateRoom(p *pubsub.V2InvalidateRoom) {
...
+		hub.WithScope(func(scope *sentry.Scope) {
+			scope.SetContext(internal.SentryCtxKey, map[string]any{
+				"room_id": p.RoomID,
+			})
+			hub.CaptureException(err)
+		})
+		logger.Err(err).
+			Str("room_id", p.RoomID).
+			Msg("Failed to fetch members after cache invalidation")
+	}
```
kegsay, line 859: Unclear if fall through is intentional here. Comment or `return` please.

sync3/handler/handler.go
```diff
@@ -818,7 +834,46 @@ func (h *SyncLiveHandler) OnInvalidateRoom(p *pubsub.V2InvalidateRoom) {
...
+
+	h.Dispatcher.OnInvalidateRoom(p.RoomID, joins, invites)
+
+	// 3. Destroy involved users' caches.
+	for _, userID := range involvedUsers {
```
kegsay, line 864: `involvedUsers` is potentially very large (think Matrix HQ). It would be better to cut this down to the connected users first, then continue. Otherwise we will spin a lot of cycles/locks on users not even on the upstream HS...

kegsay (approved):

state/accumulator.go
DMRobertson: 9ad93b8

sync3/handler/handler.go
DMRobertson: Unintentional. 041965f

sync3/handler/handler.go
DMRobertson: > It would be better to cut this down to the connected users first, then continue.

Does this run the risk of introducing a race?

sync2/handler2/handler.go
DMRobertson: They are not mutually exclusive (though I think the question you mean to ask is "are they orthogonal?").

sync3/handler/handler.go
DMRobertson: PTAL at 78b1e59 and 4011e38.

DMRobertson: And 9aa8f55 :facepalm:

sync2/handler2/handler.go
DMRobertson: What I really want is a sum type to be able to express three distinct states:

```
Noop
NewSnapshot(int64 snapID)
ReplacedSnapshot(int64 newSnapID)
```

I'm tempted to leave this as it is tbh.

kegsay (approved):
Please do not merge this yet. I want to land #367 first as a safe deploy before trying a riskier bigger change.

Issue #365, opened 2023-11-06
Messages between invite/join that should be readable get dropped, with no way for clients to backfill


kegsay:
When working on E2EE tests, I noticed that the SS proxy response when transitioning from invite -> join, with a message inbetween, is wrong. Specifically, the response ends up looking like:
```json
{
    "lists": {
        "a": {
            "count": 1
        }
    },
    "rooms": {
        "!tNyeXzIPyYnCIQfVvi:synapse": {
            "name": "user-1699267183-1",
            "avatar": null,
            "timeline": [
                {
                    "type": "m.room.member",
                    "sender": "@user-1699267183-1:synapse",
                    "content": {
                        "membership": "invite",
                        "displayname": "user-1699267183-2"
                    },
                    "state_key": "@user-1699267183-2:synapse",
                    "origin_server_ts": 1699267183520,
                    "unsigned": {
                        "age": 151
                    },
                    "event_id": "$x21JkRd2NLTbQvy0safVktBXbklAZNyTKQKpCUlLyA4"
                },
                {
                    "type": "m.room.member",
                    "sender": "@user-1699267183-2:synapse",
                    "content": {
                        "membership": "join",
                        "displayname": "user-1699267183-2"
                    },
                    "state_key": "@user-1699267183-2:synapse",
                    "origin_server_ts": 1699267183891,
                    "unsigned": {
                        "replaces_state": "$x21JkRd2NLTbQvy0safVktBXbklAZNyTKQKpCUlLyA4",
                        "prev_content": {
                            "membership": "invite",
                            "displayname": "user-1699267183-2"
                        },
                        "prev_sender": "@user-1699267183-1:synapse",
                        "age": 88
                    },
                    "event_id": "$ocxBT8dz9I4BUkDk1gU--v2sTpgk9ILFftPL6mBN3vY"
                }
            ],
            "notification_count": 0,
            "highlight_count": 0,
            "initial": true,
            "joined_count": 2,
            "invited_count": 0,
            "prev_batch": "t7-33_4_0_1_1_1_1_9_0_1",
            "num_live": 1,
            "timestamp": 1699267183891
        }
    },
    "extensions": {},
    "pos": "2"
}
```

It's wrong because it is omitting the message inbetween invite/join. Correct responses would either:
 - Return the join event with a `prev_batch` token to pass to `/messages` (if the proxy wanted to ignore history visibility calcs)
[...]

PR #366, opened 2023-11-06
Modify which events get returned when requesting initial timeline events

kegsay:
As outlined in #365, we were trying to be too smart before by calculating all the ranges for the room and returning them as a slice of from/to positions. This neglected to consider when history visibility was set to world_readable/shared/invited. In those scenarios, if the invite and join were in close proximity in the timeline (which is fairly typical), such that the `timeline_limit` would encompass both the invite and the join, the calculated timeline would incorrectly omit the events between the invite and the join. This is very bad, because it means the client has no way to fetch those messages using `/messages`, as the `prev_batch` token will relate to the _invite_ (`timeline[0]`).

This PR fixes this by only returning a single range, the last joined range for that user. This ensure the prev_batch token is always correct, so the client can hit `/messages` with it and the upstream HS can calculate history visibility.

It's worth noting that there isn't much of a tradeoff here. The old behaviour was worse in almost all circumstances. It was only "better" in the rare case of:
 - the room has `joined` history visibility
 - the timeline limit encompasses multiple join/leave memberships for the connecting user.

In that one case, the proxy would be able to return more timeline events correctly, thus potentially cutting the need for a `/messages` hit.

It looks like this bug has sat in the codebase since the beginning, and whilst we do have membership transition tests, and we did know we were conservative on history visibility checks, we didn't combine the two in any tests to check that clients can get the data they need. This PR also adds a regression test which explicitly hits `/messages` with the provided `prev_batch` token to ensure that clients see the right data.

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"strings"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/matrix-org/sliding-sync/sync2"
 
 	"github.com/getsentry/sentry-go"
@@ -716,17 +718,18 @@ func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64,
 			if err != nil {
 				return fmt.Errorf("room %s failed to SelectEventsBetween: %s", roomID, err)
 			}
-			// keep pushing to the front so we end up with A,B,C
 			for _, ev := range events {
 				if latestEventNID == 0 { // set first time and never again
 					latestEventNID = ev.NID
 				}
-				roomEvents = append([]json.RawMessage{ev.JSON}, roomEvents...)
+				roomEvents = append(roomEvents, ev.JSON)
 				earliestEventNID = ev.NID
 				if len(roomEvents) >= limit {
 					break
 				}
 			}
+			// we want the most recent event to be last, so reverse the slice now in-place.
+			slices.Reverse(roomEvents)
 			latestEvents := LatestEvents{
 				LatestNID: latestEventNID,
 				Timeline:  roomEvents,
@@ -747,8 +750,8 @@ func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64,
 }
 
 // visibleEventNIDsBetweenForRooms determines which events a given user has permission to see.
-// It accepts a nid range [from, to]. For each given room, it calculates the NID ranges
-// [A1, B1], [A2, B2], ... within [from, to] in which the user has permission to see events.
+// It accepts a nid range [from, to]. For each given room, it calculates the NID range
+// [A1, B1] within [from, to] in which the user has permission to see events.
 func (s *Storage) visibleEventNIDsBetweenForRooms(userID string, roomIDs []string, from, to int64) (map[string][2]int64, error) {
 	// load *THESE* joined rooms for this user at from (inclusive)
 	var membershipEvents []Event
--- a/state/storage.go
+++ b/state/storage.go
@@ -701,37 +701,30 @@ func (s *Storage) RoomStateAfterEventPosition(ctx context.Context, roomIDs []str
 // - with NIDs <= `to`.
 // Up to `limit` events are chosen per room.
 func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64, limit int) (map[string]*LatestEvents, error) {
-	roomIDToRanges, err := s.visibleEventNIDsBetweenForRooms(userID, roomIDs, 0, to)
+	roomIDToRange, err := s.visibleEventNIDsBetweenForRooms(userID, roomIDs, 0, to)
 	if err != nil {
 		return nil, err
 	}
 	result := make(map[string]*LatestEvents, len(roomIDs))
 	err = sqlutil.WithTransaction(s.Accumulator.db, func(txn *sqlx.Tx) error {
-		for roomID, ranges := range roomIDToRanges {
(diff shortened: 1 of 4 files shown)


state/storage.go
```diff
@@ -701,38 +701,31 @@ func (s *Storage) RoomStateAfterEventPosition(ctx context.Context, roomIDs []str
...
+			events, err := s.EventsTable.SelectLatestEventsBetween(txn, roomID, r[0]-1, r[1], limit)
+			if err != nil {
+				return fmt.Errorf("room %s failed to SelectEventsBetween: %s", roomID, err)
+			}
+			// keep pushing to the front so we end up with A,B,C
+			for _, ev := range events {
+				if latestEventNID == 0 { // set first time and never again
+					latestEventNID = ev.NID
+				}
+				roomEvents = append([]json.RawMessage{ev.JSON}, roomEvents...)
```
DMRobertson, line 724: Aside: is this performant? I would expect that this ends up making a new backing array once for each iteration.

state/storage.go
```diff
@@ -756,7 +749,7 @@ func (s *Storage) LatestEventsInRooms(userID string, roomIDs []string, to int64,
 // visibleEventNIDsBetweenForRooms determines which events a given user has permission to see.
 // It accepts a nid range [from, to]. For each given room, it calculates the NID ranges
 // [A1, B1], [A2, B2], ... within [from, to] in which the user has permission to see events.
```
DMRobertson, line 751: Can we update this to say that we only return one NID range, please?

tests-e2e/timeline_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/matrix-org/complement/b"
+	"github.com/matrix-org/complement/must"
+	"github.com/matrix-org/sliding-sync/sync3"
+	"github.com/matrix-org/sliding-sync/testutils/m"
+)
+
+// Regression test to make sure that given:
+// - Alice invite Bob
+// - Alice send message
+// - Bob join room
```
DMRobertson, line 16: Can we spell out the assumptions on history visibility here please?

state/storage.go
kegsay: Yes it will do. So no, it's churny on memory. We should probably just go to the back then in-place reverse? Reducing memory spikes would be useful.

state/storage.go
kegsay: Bah, forgot that comment :D

state/storage.go
DMRobertson: Don't think we need to do it here. If the limit is small (~20?) then this is probably minimal?

Could also
-  `make([]json.RawMessage, len(events))` 
- write starting at index `len(events)-1` and decrease?

(Tbh I'm surprise we don't get the DB to return events in ascending NID order)

state/storage.go
```diff
@@ -879,11 +873,16 @@ func (s *Storage) visibleEventNIDsWithData(joinTimingsAtFromByRoomID map[string]
 		if isJoined {
 			result = append(result, [2]int64{startIndex, toIncl})
 		}
-		return result
+		if len(result) == 0 {
+			return [2]int64{}
+		}
+		// we only care about the LAST nid range, otherwise we can end up with gaps being returned in the
+		// timeline. See https://github.com/matrix-org/sliding-sync/issues/365
```
DMRobertson, line 880: I think this wording is going to confuse my future self. (Ditto for the other comment citing this issue). It makes it sound like "gaps in the timeline" is fundamentally bad---but we _do_ want there to be events not sent to users in certain situations (e.g. `joined` history vis).

"Gaps in the timeline" means presumably means(?): the proxy returns a timeline [A, ...., B, C, ... D] such that
1. B is not a direct ancestor of C,
2. there is some event between B and C that the user has permission to see, and
3. there is nothing in the final sync response which tells clients about points (1) and (2)?

Maybe we can say "otherwise we can end up with _incorrect_ gaps in the timeline"?

state/storage.go
kegsay: The problem is this is done per room, so if you have a 50 window, then 50x`timeline_limit` events (worst case).

kegsay: Using `slices.Reverse` now.

DMRobertson (approved):

state/storage.go
kegsay: I dunno, for me this seems clear. A "gap" fundamentally is in between two things. In this case, there is a gap between two timeline events e.g:
```
[A,B,C,D,E] true timeline
[A,B,D,E]   calculated timeline with a gap
```

Hence, "gaps in the timeline" _are_ fundamentally bad, as it means we have omitted an event.

Technically, we could be pedantic and say that a "gap" is also present when you have:
```
[A,B,Leave, .... , Join, D, E]
```

and that this is a "correct" gap (assuming joined history visibility), but this has a ton of caveats and ultimately isn't something the proxy should ever do anymore as of this PR. The moment we start differentiating between correct (which won't be producible by the proxy) and incorrect gaps, it just adds more ultimately unnecessary lexicon imo.

This is further supported by how end-users describe it in bug reports:
 - "I have gap in my timeline"
 - "There are gaps in my timeline"
 - "Sent a message from my matrix.org account. Got notification on EIX, opened it from there. When the timeline loaded, there is a gap in the timeline"

Let's use the same term to mean the same thing, a bad thing.

DMRobertson: it also doesn't help that we speak of "gappy" v2 syncs etc.

PR #369, opened 2023-11-08
Migrate stuck invites

kegsay:
With tests.

### DIFF
--- a/state/migrations/20231108122539_clear_stuck_invites.go
+++ b/state/migrations/20231108122539_clear_stuck_invites.go
@@ -33,7 +33,7 @@ func upClearStuckInvites(ctx context.Context, tx *sql.Tx) error {
 		WHERE room_id NOT IN (
 			SELECT room_id
 			FROM syncv3_rooms
-		) GROUP BY user_id
+		)
 	`)
 	defer rows.Close()
 	if err != nil {
--- a/state/migrations/20231108122539_clear_stuck_invites.go
+++ b/state/migrations/20231108122539_clear_stuck_invites.go
@@ -1,0 +1,83 @@
+package migrations
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/pressly/goose/v3"
+)
+
+func init() {
+	goose.AddMigrationContext(upClearStuckInvites, downClearStuckInvites)
+}
+
+// The purpose of this migration is to find users who have rooms which have
+// not been properly processed by the proxy and invalidate their since token
+// so they will do an initial sync on the next poller startup. This is specifically
+// targeting stuck invites, where there is an invite in the invites table but
+// the room is already joined. This is usually (always?) due to missing a create
+// event when the room was joined, caused by a synapse bug outlined in
+// https://github.com/matrix-org/sliding-sync/issues/367
+// This isn't exclusively a problem with invites, though it manifests more clearly there.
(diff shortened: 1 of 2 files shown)


state/migrations/20231108122539_clear_stuck_invites.go
```diff
@@ -0,0 +1,83 @@
...
+	// room into rooms.join"? For every room in rooms.join, we should have seen a create event for it, and hence have
+	// an entry in syncv3_rooms. If we do not have an entry in syncv3_rooms but do have an entry in syncv3_unread, this
+	// implies we failed to properly store this joined room and therefore the user who the unread marker is for should be
+	// reset to force an initial sync. On matrix.org, of the users using sliding sync, this will catch around ~1.82% of users
+	rows, err := tx.QueryContext(ctx, `
+		SELECT distinct(user_id) FROM syncv3_unread
+		WHERE room_id NOT IN (
+			SELECT room_id
+			FROM syncv3_rooms
+		) GROUP BY user_id
```
DMRobertson, line 36: Is the GROUP BY needed here, given that we're already using DISTINCT?

state/migrations/20231108122539_clear_stuck_invites_test.go
```diff
@@ -0,0 +1,212 @@
...
+	// Test cases:
+	// Room | In Invite Table? | In Unread Table? | In Room Table? | Comment
+	//   A           Y                 Y                Y             OK, Genuine invite, proxy in room
+	//   B           Y                 Y                N             BAD, Stuck invite, proxy never saw join
+	//   C           Y                 N                Y             OK, Genuine invite, proxy in room, no unread counts (unusual but valid)
+	//   D           Y                 N                N             OK, Genuine invite, proxy not in room
+	//   E           N                 Y                Y             OK, Genuine joined room
+	//   F           N                 Y                N             BAD, Stuck joined room, proxy never saw join
+	//   G           N                 N                Y             OK, Genuine joined room, no unread counts (unusual but valid)
+	//   -           N                 N                N             Impossible, room id isn't in any table!
```
DMRobertson, line 52: There's an assumption here that we never delete from the rooms table. Is that true? (I think it should be, but let's check)

state/migrations/20231108122539_clear_stuck_invites_test.go
```diff
@@ -0,0 +1,212 @@
...
+	// invites for Bob and Doris are gone
+	for _, userID := range []string{bob, doris} {
+		got, err := inviteTable.SelectAllInvitesForUser(userID)
+		if err != nil {
+			t.Fatalf("SelectAllInvitesForUser: %s", err)
+		}
+		if len(got) > 0 {
+			t.Fatalf("SelectAllInvitesForUser got invites for %s, wanted none: %+v", userID, got)
+		}
+	}
```
DMRobertson, line 211: Can we check that the genuine invites remain too?

DMRobertson (approved):

state/migrations/20231108122539_clear_stuck_invites_test.go
kegsay: Yes, with the exception of a migration which did delete "bad" rooms.

state/migrations/20231108122539_clear_stuck_invites.go
kegsay: No it isn't.

Issue #110, opened 2023-05-12
The prev_batch value for room responses can be inaccurate


kegsay:
`prev_batch` is a token which can be passed to `/messages` to collect more data. It is sent by the upstream HS and the proxy cannot fudge/fake them. This causes problems in the following situation:
 - v2 poller gets 10 timeline events and a single prev_batch in a single response.
 - the events and prev_batch are persisted in the proxy's DB.
 - the sliding sync client asks for a timeline limit of 7.
 - what value do we return for the prev_batch?

The possible solutions are:
 - If we return the prev_batch we got from v2, the client will never see the first 3 events in the batch of 10, because the `prev_batch` starts at the earliest event, not the latest.
 - If we return no prev_batch then the client cannot backpaginate at all.
 - We could return a LATER prev_batch (if one exists), but this would cause duplicate events. <-- we currently do this.

The current solution is suboptimal due to the duplication of events and the inherent unreliability of the token (it may not exist depending on the precise limit used in the client + responses received by the HS). 

A native HS impl can actually calculate these values so does not face this problem.

PR #377, opened 2023-11-14
Add cache invalidation tests (rejoining rooms)

kegsay:
Fixes two things:
 - refreshing the avatar URL didn't work because it read the wrong field in `content`.
 - live timeline events did not include a `prev_batch` field which is vital when `prev_batch` is missing due to proxy limitations (#110)

### DIFF
--- a/state/storage.go
+++ b/state/storage.go
@@ -316,7 +316,7 @@ func (s *Storage) ResetMetadataState(metadata *internal.RoomMetadata) error {
 	FROM syncv3_events JOIN snapshot ON (
 		event_nid = ANY (ARRAY_CAT(events, membership_events))
 	)
-	WHERE (event_type IN ('m.room.name', 'm.room.avatar', 'm.room.canonical_alias') AND state_key = '')
+	WHERE (event_type IN ('m.room.name', 'm.room.avatar', 'm.room.canonical_alias', 'm.room.encryption') AND state_key = '')
 	   OR (event_type = 'm.room.member' AND membership IN ('join', '_join', 'invite', '_invite'))
 	ORDER BY event_nid ASC
 	;`, metadata.RoomID)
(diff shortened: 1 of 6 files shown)


state/storage.go
```diff
@@ -365,7 +367,7 @@ func (s *Storage) ResetMetadataState(metadata *internal.RoomMetadata) error {
 		metadata.Heroes = append(metadata.Heroes, hero)
 	}
 
-	// For now, don't bother reloading Encrypted, PredecessorID and UpgradedRoomID.
+	// For now, don't bother reloading PredecessorID and UpgradedRoomID.
 	// These shouldn't be changing during a room's lifetime in normal operation.
```
DMRobertson, line 371: For completeness: two users could emit tombstones concurrently, and then the winner changes due to state res. But I chose to ignore that when writing this comment.

tests-e2e/state_updates_test.go
```diff
@@ -0,0 +1,251 @@
...
+				BumpEventTypes: []string{"m.room.message"},
+			},
+		},
+	})
+	m.MatchResponse(t, res, m.MatchRoomSubscriptionsStrict(
+		map[string][]m.RoomMatcher{
+			roomID: {
+				m.MatchJoinCount(2),
+				m.MatchRoomName(bob.Localpart),
+				// despite asking for 5 timeline events, we only get 1 on the initial sync so prev_batch is correct
```
DMRobertson, line 52: Confused by this. I would have thought that Alice's timeline at least includes her join and her message? If it did, what would stop us from giving a correct prev_batch token?

tests-e2e/state_updates_test.go
```diff
@@ -0,0 +1,251 @@
...
+				Ranges:         sync3.SliceRanges{{0, 20}},
+				BumpEventTypes: []string{"m.room.message"},
+			},
+		},
+	}, roomID, aliceJoin)
+	m.MatchResponse(t, res, m.MatchRoomSubscriptionsStrict(map[string][]m.RoomMatcher{
+		roomID: {
+			m.MatchInviteCount(1), // doris
+			m.MatchJoinCount(2),   // alice and charlie
+			MatchRoomTimeline([]Event{aliceJoin}),
```
DMRobertson, line 139: Why not Charlie's messages?

DMRobertson (approved):

tests-e2e/state_updates_test.go
kegsay: This request is an _initial_ sync v2 request. Initial sync v2 requests always use a v2 filter of timeline limit = 1, to ensure we can set a valid `prev_batch` token. This is because if we instead did 50 events and the client asked for 5 events, we have no prev_batch at event 5,4,3,2,1 which we could pass to the user. Will clarify, as I think because I neglected to emphasise "initial" as in initial v2 req, it's not clear.

tests-e2e/state_updates_test.go
kegsay: Because alice's join is the latest event, and the proxy always applies "joined" history visibility, relying on the client hitting /messages for scrollback, hence it's bounded by her join.

Issue #216, opened 2023-07-21
[Performance] Implement Unix domain socket


nisbet-hubbard:
This probably won’t be a priority. That said, Unix domain socket is known to perform better than TCP socket, and is considered more secure. It also seems to be the direction homeserver implementations are heading.

It could be offered as an option for setups where the sliding sync proxy lives on the same machine as the reverse proxy.

kegsay:
This is done now.

PR #378, opened 2023-11-16
unix socket support

cyberb:
Allows using unix socket on both ends:

```
SYNCV3_BINDADDR=/var/snap/matrix/current/sliding-sync.socket
SYNCV3_SERVER=/var/snap/matrix/current/matrix.socket
```

Tested in Syncloud (https://github.com/syncloud/matrix) which allows me to use Element X now.

### Pull Request Checklist


sync2/client.go
```diff
@@ -39,16 +41,30 @@ type HTTPClient struct {
...
-		DestinationServer: destHomeServer,
+		LongTimeoutClient: newClient(longTimeout, destHomeServer),
+		Client:            newClient(shortTimeout, destHomeServer),
+		DestinationServer: baseUrl,
+	}
+}
+
+func newClient(timeout time.Duration, destHomeServer string) *http.Client {
+	transport := http.DefaultTransport
+	if strings.HasPrefix(destHomeServer, "/") {
```
kegsay, line 58: Fairly sure unix sockets can also be relative file paths, so this isn't ideal (albeit it is convenient). I would definitely split this out to a helper function in case we want to be a bit more intelligent around unix socket detection. E.g something in `internal/util.go`:
```go
func isUnixSocket(httpOrUnixStr string) bool {
    return strings.HasPrefix(httpOrUnixStr, "/")
}
```

It might be better to just say `strings.HasPrefix("http")`? Not really sure.

sync2/client.go
```diff
@@ -39,16 +41,30 @@ type HTTPClient struct {
...
+		transport = &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", destHomeServer)
```
kegsay, line 61: I dislike how unintuitive this is. Initially, we say "if it's unix, set the HS URL to `http://unix`" but the real magic comes from clobbering the transport function to dial the real file. Any poor developer reading this could easily think `destHomeServer == "http://unix"` and be very confused at this line.

v3.go
```diff
@@ -216,12 +219,18 @@ func RunSyncV3Server(h http.Handler, bindAddr, destV2Server, tlsCert, tlsKey str
 
 	// Block forever
 	var err error
-	if tlsCert != "" && tlsKey != "" {
-		logger.Info().Msgf("listening TLS on %s", bindAddr)
-		err = http.ListenAndServeTLS(bindAddr, tlsCert, tlsKey, srv)
+	if strings.HasPrefix(bindAddr, "/") {
```
kegsay, line 222: Docs need updating to mention that `SYNCV3_BINDADDR` can now be a unix socket. Same with `SYNCV3_SERVER`.

kegsay (changes requested):
Overall I'm happy with the idea, but there's some extra bits to do to get it over the line:
 - Factor out the unix socket detection to a helper function (aids readability and allows us to be more intelligent easily in the future)
 - Mention this new support in the docs (README and the big help message in `./cmd/syncv3/main.go`)
 - Run E2E tests with unix sockets so we can make sure this works and won't bitrot.

Thanks for your contribution!

sync2/client.go
cyberb: extracted HomeServerUrl is it better?

cyberb:
e2e test use real synapse server which I am not sure even supports unix sockets (dendrite does), I will need to check if there is an env, but even if there is do you want to run all e2e tests on each type (http and unix)?
I would just add one test to check that unix socket client can make requests.

kegsay:
Hmm, good point. Synapse _does_ support unix sockets via `SYNAPSE_USE_UNIX_SOCKET=1` in the synapse service image we're using, but I'm not sure you can talk unix sockets to the GHA service container. Running a single test will do, though it isn't super ideal as I'm interested in testing large packet sizes / load.

Let me know how much work this will be, and if it's not really feasible we'll think of something else.

internal/errors.go
```diff
@@ -90,13 +90,13 @@ func Assert(msg string, expr bool, debugContext ...map[string]interface{}) {
...
 func AssertWithContext(ctx context.Context, msg string, expr bool) {
-	assert(msg, expr)
+	assertCustom(msg, expr)
 	if !expr {
 		GetSentryHubFromContextOrDefault(ctx).CaptureException(fmt.Errorf("assertion failed: %s", msg))
 	}
 }
 
-func assert(msg string, expr bool) {
+func assertCustom(msg string, expr bool) {
```
kegsay, line 99: Why the rename?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/pressly/goose/v3 v3.14.0
 	github.com/prometheus/client_golang v1.13.0
 	github.com/rs/zerolog v1.29.0
+	github.com/stretchr/testify v1.8.4
```
kegsay, line 18: Please don't add an entire test package just for basic equality checks.

internal/home_server_url.go
```diff
@@ -0,0 +1,25 @@
+package internal
+
+import "strings"
+
+type HomeServerUrl struct {
+	HttpOrUnixStr string
+}
+
+func (u HomeServerUrl) IsUnixSocket() bool {
```
kegsay, line 9: This isn't what I was getting at before. It would be quite enough to just have:
```go
func IsUnixSocket(u string) bool {
    return strings.HasPrefix(u, "/")
}
```
This is better than before because then the implementation of prefix checking is only in a single place, so if we want to swap to something better-than-prefix-checking we just need to edit one place, and it reads more nicely and self-documents:
```go
if internal.IsUnixSocket(destHomeserverUrl) {
    // set up unix transport
}
```

We don't need the boilerplate struct and getters here, as they don't add any value. They don't add value because the caller still needs to care about the response from this function to determine if the call to `GetUnixSocket()` or `GetBaseUrl()` would make any sense, so the caller's code would look a lot like the above, just with a struct definition instead of a simple static function. If we were going to go down the struct road, it would be better to have a way to map from a `HttpOrUnixStr` to an `http.Transport`, which could be called without caring about whether this is a unix or HTTP string.

internal/home_server_url.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func (u HomeServerUrl) GetUnixSocket() string {
+	if u.IsUnixSocket() {
+		return u.HttpOrUnixStr
+	}
+	return ""
+}
+
+func (u HomeServerUrl) GetBaseUrl() string {
```
kegsay, line 20: Idiomatic Go capitalises acronyms so `GetBaseURL` and `HomeServerURL`.

v3.go
```diff
@@ -230,6 +239,22 @@ func RunSyncV3Server(h http.Handler, bindAddr, destV2Server, tlsCert, tlsKey str
...
+func unixSocketListener(bindAddr string) net.Listener {
+	err := os.Remove(bindAddr)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		logger.Fatal().Err(err).Msg("failed to remove existing unix socket")
+	}
+	listener, err := net.Listen("unix", bindAddr)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("failed to serve unix socket")
+	}
+	err = os.Chmod(bindAddr, 0755)
```
kegsay, line 251: Please document the respective r/w/x permissions you wish to give to this socket as a comment.

kegsay (changes requested):
I don't think `HomeServerUrl` makes things much clearer here, and it adds boilerplate where it is not necessary. Please:
 - revert it back and just add `internal.IsUnixSocket(u string) bool` and call that in places where you previously had `strings.HasPrefix`. That is just as easily unit testable should you wish to do so, but please do not add `testify` for basic equality checks.
 - add a function to map a unix socket string to a unix transport: `internal.UnixTransport(u string) http.Transport`
 - tie the two functions together in `newClient`:
```go
func newClient(timeout time.Duration, destinationURL string) *http.Client {
    transport := http.DefaultTransport
    if internal.IsUnixSocket(destinationURL) (
        transport = internal.UnixTransport(destinationURL)
    }
    return &http.Client{
        Timeout: timeout,
        Transport: transport,
    }
}
```

go.mod
kegsay: `go mod tidy` please.

internal/errors.go
kegsay: Please revert.

kegsay (commented):
Thanks, just a few final minor things:
 - `go mod tidy` to remove the testify dep.
 - Please document the respective r/w/x permissions you wish to give to this socket as a comment.
 - Revert the change from `assert` to `assertCustom`.

Then LGTM, thanks!

kegsay (approved):

cyberb:
Thank you!
Probably this issue can be closed as well: https://github.com/matrix-org/sliding-sync/issues/216

v3.go
```diff
@@ -230,6 +239,23 @@ func RunSyncV3Server(h http.Handler, bindAddr, destV2Server, tlsCert, tlsKey str
...
+	err := os.Remove(bindAddr)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		logger.Fatal().Err(err).Msg("failed to remove existing unix socket")
+	}
+	listener, err := net.Listen("unix", bindAddr)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("failed to serve unix socket")
+	}
+	// TODO: safe default for now (rwxr-xr-x), could be extracted as env variable if needed
+	err = os.Chmod(bindAddr, 0755)
```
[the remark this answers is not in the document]
cyberb: I was under impression that 755 will only allow the owner to use it.
After googling looks like eXecute is not need led but read and weite are needed to use it.
Ideally we pass it as a parameter, if not then probably 600 is what I wanted.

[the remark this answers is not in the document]
n3vu0r: According to [unix(7)](https://man7.org/linux/man-pages/man7/unix.7.html), write permission is required to connect to a unix domain socket.

> On Linux, connecting to a stream socket object requires write permission on that socket; sending a datagram to a datagram socket likewise requires write permission on that socket.

I'd suggest to make it at least user and group connectable, that is `0660` (execution bit has no effect).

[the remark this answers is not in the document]
emilylange: First and foremost: Sorry for my original comment.
I was in a bit of a rush when I wrote it, and should just have waited a bit longer before commenting.

Multiple /bit flips/ happened in my original comment, which I want to point out:

1. `write` perms are needed to connect and talk over a unix socket, not `read`, as @n3vu0r correctly pointed out.
2. `0755` does not translate to `u=rwx,go=rw` but rather `u=rwx,go=rx`.
3. Which in turn makes, as intended, the socket not world-connectable but rather only by the owner itself.

I just was very happy to see unix socket support implemented so soon, but then saw some unusual permission bits.

Given I worked on unix socket permission bits as part of https://github.com/caddyserver/caddy/pull/4741, I figured I should comment. Which then lead to that hastily, and more importantly, incorrect comment. Sorry for that.


According to the current `unix(7)`, only `write` perms are needed.
There are, however, still a lot of old `unix(7)` versions hosted on the web, which still mention `read` and `write` being needed.

For more details on this, see https://github.com/caddyserver/caddy/pull/4741#issuecomment-1177951257.

---

> I'd suggest to make it at least user and group connectable, that is `0660` (execution bit has no effect).

I would suggest the same, but without `read`, so `0220`.

PR #384, opened 2023-12-05
WIP: only emit avatars when the DM flag is set

DMRobertson:
C.f. https://github.com/vector-im/element-x-ios/issues/2003#issuecomment-1840438661

I don't think this works yet because clients have to set the DM flag after they see the room down /sync. And the proxy probably doesn't recalculate avatars after the DM flag changes.

More generally there's a lot here that's confusing and I don't really like this code.


kegsay:
https://github.com/matrix-org/sliding-sync/pull/394

Issue #383, opened 2023-12-03
SYNCV3_BINDADDR variable for unix socket


MrSpadduzza:
Could you elaborate about this, please?
SYNCV3_BINDADDR   Default: 0.0.0.0:8008. **The interface and port to listen on. (Supports unix socket: /path/to/socket)**

I have tried various syntax versions, but I keep getting the following error:
listening on unix://etc/matrix-synapse/sliding-sync-0.99.12/
failed to listen and serve error="listen tcp: lookup tcp///etc/matrix-synapse/sliding-sync-0.99.12/: Servname not supported for ai_socktype"

The first line, where it declare the path vary according to the variable, but the second line error is always the same.

The error changes, if I omit 'unix:' from the variable path:
listening on /etc/matrix-synapse/sliding-sync-0.99.12/syncv3.socket
failed to listen and serve error="listen tcp: address /etc/matrix-synapse/sliding-sync-0.99.12/syncv3.socket: missing port in address"

I have tried with or without socket file name; the socket file does not exist and the locations user owner is the same that runs the proxy.

Any idea?

DMRobertson:
See https://github.com/matrix-org/sliding-sync/releases/tag/v0.99.13. Please try again with the latest release!

MrSpadduzza:
Thank you very much for your quick reply!! I'm trying it now and get back

MrSpadduzza:
I confirm the proxy started happily with SYNCV3_BINDADDR and SYNCV3_SERVER variables set as =/path/to/socket/file
I'll go on with some testing.
Thanks again!!

PR #386, opened 2023-12-13
Add SYNCV3_SERVER_PUB environment variable pointing to the public homeserver URL

spantaleev:
`SYNCV3_SERVER` continues to be used for connecting to the server.

`SYNCV3_SERVER_PUB` is a new environment variable (defaulting to the value of `SYNCV3_SERVER`), so that people who are already pointing `SYNCV3_SERVER` to a public URL do not need to set both.

This change affects the `server` field in the generated `/client/server.json` file, so that this field points to the public homeserver URL (`SYNCV3_SERVER_PUB`).

Related to https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/2583

### Pull Request Checklist

### DIFF
--- a/cmd/syncv3/main.go
+++ b/cmd/syncv3/main.go
@@ -36,9 +36,10 @@ var (
 
 const (
 	// Required fields
-	EnvServer = "SYNCV3_SERVER"
-	EnvDB     = "SYNCV3_DB"
-	EnvSecret = "SYNCV3_SECRET"
+	EnvServer    = "SYNCV3_SERVER"
+	EnvServerPub = "SYNCV3_SERVER_PUB"
+	EnvDB        = "SYNCV3_DB"
+	EnvSecret    = "SYNCV3_SECRET"
 
 	// Optional fields
 	EnvBindAddr               = "SYNCV3_BINDADDR"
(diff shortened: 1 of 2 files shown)


spantaleev:
I'm not sure if this is the right place to fix this.

This is to benefit Element-web, which seems to be hitting `/client/server.json` and complaining that the internal homeserver URL there does not match the homeserver it's talking to, as desribed in https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/2583

Element X (at least on iOS) does not seem to be doing such checks and does not care what `/client/server.json` says.

So it may be that element-web needs to stop caring too, or that Element X needs to be fixed. In any case, it seems like `/client/server.json` may also benefit from advertising the correct public homeserver URL anyway.

-----

Working on this, I've also noticed that https://github.com/matrix-org/sliding-sync/pull/378 (ae73ace4a41be24ebc9ca7752346659061a38338) which introduced UNIX socket support (to `SYNCV3_SERVER` and `https://github.com/matrix-org/sliding-sync/issues/383`) may also benefit from this current patch landing.

When `SYNCV3_SERVER` is pointed to a UNIX socket, it also ends up in the `/client/server.json` file (`"server": "/socket.sock", ..}` which I suppose clients like Element will not like. People using UNIX sockets can use the new `SYNCV3_SERVER_PUB` environment variable for the public homeserver URL.

I also believe that validation checks in `cmd/syncv3/main.go` should be adjusted, so that when `SYNCV3_SERVER` is pointed to a socket file, `SYNCV3_SERVER_PUB` would be required.

If you agree with this, I could prepare another PR.. Or feel free to quickly fix it up yourselves.

kegsay:
> So it may be that element-web needs to stop caring too

This. `/client/server.json` isn't part of the MSC, and element-web shouldn't care about this nowadays. The `SYNCV3_SERVER` env var is an internal URL, and really we probably shouldn't be leaking it outside (via `/client/server.json`). Adding another env var just confuses things more.

spantaleev:
Sounds good!

I wonder why element-web is trying to read it from there though. Perhaps that should be fixed first.

kegsay:
Nothing to do here, so closing. Element-Web should not be checking `/client/server.json` anymore.

PR #393, opened 2024-01-13
CGO disabled to allow running on older distros

csett86:
with CGO we are linking to whatever glibc is in ubuntu-latest, which may
not be present in other distros (Think older Debian, Alpine)

In addition, the docker build with alpine already (implicitly) does not
use CGO, as CGO is still not supported with musl officially if I
understand the following issue correctly: https://github.com/golang/go/issues/19938

So disabling CGO in the released binaries and in the README instructions harmonizes
the behaviour across docker and non-docker builds.

### Pull Request Checklist

Signed-off-by: Christoph Settgast <csett86_git@quicksands.de>


kegsay (approved):
Seems sensible, thanks!

PR #394, opened 2024-01-22
Only emit avatars when DM room is set

kegsay:
Based on https://github.com/matrix-org/sliding-sync/pull/384 with modifications.


sync3/caches/user.go
```diff
@@ -115,7 +115,7 @@ func NewInviteData(ctx context.Context, userID, roomID string, inviteState []jso
 					Timestamp:     uint64(ts),
 					AlwaysProcess: true,
 				}
-				id.IsDM = j.Get("is_direct").Bool()
+				id.IsDM = j.Get("content.is_direct").Bool()
```
kegsay, line 118: This was just a general bug.

S7evinK (approved):

PR #418, opened 2024-04-11
bugfix: don't send dupe room account data

### LINKED ISSUE #417: Room account data can be spammed with duplicates
Reported by @bnjbvr 

I think the account data extension can send room account data multiple times redundantly when catching up on a nearly full buffer. E.g if you have 100 items for the same room in the buffer, the account data extension pulls the room account data 100 times and appends it to the response.

Close when this has a test showing it can't happen / can happen and bug is fixed.

EDIT: seen a mitmdump which shows this. Sporadic room account data are duplicated 3,8,N times. Hypothesis: If you SYNC a bunch of rooms whilst there are RoomUpdate items in the buffer, it will cause duplicates.

kegsay:
Fixes #417

### DIFF
--- a/sync3/extensions/account_data.go
+++ b/sync3/extensions/account_data.go
@@ -3,6 +3,7 @@ package extensions
 import (
 	"context"
 	"encoding/json"
+
 	"github.com/matrix-org/sliding-sync/internal"
 	"github.com/matrix-org/sliding-sync/state"
 	"github.com/matrix-org/sliding-sync/sync3/caches"
@@ -21,6 +22,8 @@ func (r *AccountDataRequest) Name() string {
 type AccountDataResponse struct {
 	Global []json.RawMessage            `json:"global,omitempty"`
 	Rooms  map[string][]json.RawMessage `json:"rooms,omitempty"`
+	// which rooms have had account data loaded from the DB in this response
+	loadedRooms map[string]bool
 }
 
 func (r *AccountDataResponse) HasData(isInitial bool) bool {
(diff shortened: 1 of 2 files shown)


sync3/extensions/account_data.go
```diff
@@ -21,6 +22,8 @@ func (r *AccountDataRequest) Name() string {
 type AccountDataResponse struct {
 	Global []json.RawMessage            `json:"global,omitempty"`
 	Rooms  map[string][]json.RawMessage `json:"rooms,omitempty"`
+	// which rooms have had account data loaded from the DB in this response
+	loadedRooms map[string]bool
```
S7evinK, line 26:
```suggestion
	loadedRooms map[string]struct{}
```
Maybe?

sync3/extensions/account_data.go
```diff
@@ -54,6 +57,13 @@ func (r *AccountDataRequest) AppendLive(ctx context.Context, res *Response, extC
 		}
 		// if this is a room update which is included in the response, send account data for this room
 		if _, exists := extCtx.RoomIDToTimeline[update.RoomID()]; exists {
+			// ..but only if we haven't before
+			if res.AccountData != nil && res.AccountData.loadedRooms[update.RoomID()] {
+				// we've loaded this room before, don't do it again
+				// this can happen when we consume lots of items in the buffer. If many of them are room updates
+				// for the same room, we could send dupe room account data if we didn't do this check.
+				break
```
S7evinK, line 65: Does it make sense to just `return` here? (same for the other break above)
Given the control flow, we would be hitting the below
```go
if len(globalMsgs) == 0 && len(roomToMsgs) == 0 {
	return
}
```

S7evinK (approved):
LGTM

sync3/extensions/account_data.go
kegsay: It's easier to read if they are bools, and we don't need the space optimisation here, so I err'd on clarity over performance.

sync3/extensions/account_data.go
kegsay: Oh, right you are! That will be clearer, thank you!

PR #419, opened 2024-04-11
Update ttlcache to v3.2.0

V02460:
Updates the `ttlcache` dependency from version 2.8.1 to 3.2.0.

Please note: I’m not especially familiar with Go, please keep this in mind while reviewing. This PR is kept a draft until I can make CI happy.

### Pull Request Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/fxamacker/cbor/v2 v2.5.0
 	github.com/getsentry/sentry-go v0.24.1
 	github.com/gorilla/mux v1.8.0
-	github.com/jellydator/ttlcache/v2 v2.11.1
+	github.com/jellydator/ttlcache/v3 v3.2.0
 	github.com/jmoiron/sqlx v1.3.3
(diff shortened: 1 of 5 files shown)


kegsay:
What is the rationale for this change?

kegsay:
I've sat on this for a while thinking about whether I want this. I've concluded that no, I don't want this. The main reasons why are:
 - changing the version introduces uncertainties around edge cases and performance, exacerbated by the breaking API changes which need code changes in the proxy.
 - the current version is not deprecated, nor is there any other pressing reason to move to the latest version e.g CVEs.

There is also sadly a part of me that is particularly wary about accepting a random unsolicited dependency bump from a first time contributor. Whilst I don't have reasons to believe there is anything nefarious here, when combined with the lack of a good justification for this change, means I'm inclined to close this.

Sorry it took so long to come to a decision on this.

V02460:
Thank you for taking the time explaining your rationale. Dismissing contributions is part of good project governance and seems very justified here. Also, be assured I don’t take personal offense in this :)

PR #424, opened 2024-04-25
Check for unstable prefix in MSC4115

kegsay:
Missed this in the original impl.

### DIFF
--- a/sync2/handler2/handler.go
+++ b/sync2/handler2/handler.go
@@ -274,6 +274,7 @@ func (h *Handler) Accumulate(ctx context.Context, userID, deviceID, roomID strin
 	for i := range timeline.Events {
 		// Delete MSC4115 field as it isn't accurate when we reuse the same event for >1 user
 		timeline.Events[i], _ = sjson.DeleteBytes(timeline.Events[i], "unsigned.membership")
+		timeline.Events[i], _ = sjson.DeleteBytes(timeline.Events[i], `unsigned.io\.element\.msc4115\.membership`)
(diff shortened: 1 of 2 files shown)


sync2/handler2/handler.go
```diff
@@ -274,6 +274,7 @@ func (h *Handler) Accumulate(ctx context.Context, userID, deviceID, roomID strin
 	for i := range timeline.Events {
 		// Delete MSC4115 field as it isn't accurate when we reuse the same event for >1 user
 		timeline.Events[i], _ = sjson.DeleteBytes(timeline.Events[i], "unsigned.membership")
+		timeline.Events[i], _ = sjson.DeleteBytes(timeline.Events[i], `unsigned.io\.element\.msc4115\.membership`)
```
devonh, line 277: Why do not all `.` need to be `\.`? (the one between `unsigned.io` isn't)
Is this specific to how synapse formats things?
Might be worth a comment explaining where these `\` come from.

devonh (approved):
I trust that this works as implemented. But I'm unsure why the `\.` is necessary.

sync2/handler2/handler.go
kegsay: Because `.` means "look at a field in this object" which makes it impossible to refer to keys which actually have "."s in them. To do this, you need to escape them e.g `foo.bar\.baz` refers to the value "quuz":
```json
{
  "foo": {
    "bar.baz": "quuz"
  }
}
```

devonh: ah I see. makes sense

PR #429, opened 2024-05-03
Finish 405 responses to avoid hanging

turt2live:
### Pull Request Checklist

- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)


turt2live:
test failure appears to be not my fault

turt2live:
this patch doesn't appear to work: it still hangs for me. I don't know why.

kegsay:
I was going to say that this should make no difference. `ServeHTTP` simply asks that you call `WriteHeader`, it doesn't mandate you set headers, read the req body or set a response body.

kegsay:
Nothing to do here.

PR #432, opened 2024-05-09
Ensure device list updates are robust to race conditions and network failures

### LINKED ISSUE #430: device lists don't get resent on rapid changes
Whilst working on https://github.com/matrix-org/complement-crypto/pull/50 I took a tcpdump of a failing test and found convincing evidence that SS is not correctly telling clients of the _later_ device list update for the _same user_. This is problematic because it means clients won't hit `/keys/query` at the right time (**after** the client has uploaded device keys) and hence won't encrypt for that device, causing UTDs.

```
alice rust = authorization: Bearer syt_dXNlci0xLWFsaWNl_pWOuJBbfdUlWnqLLVRgm_1rPA9l\r\n
bob JS = Authorization: Bearer syt_dXNlci0yLWJvYg_BCEPDhFaFLDZGIrkSuVX_1q3cQp\r\n
bob NEW_DEVICE String value: syt_dXNlci0yLWJvYg_RiemTZKJheBtUhAIxhWu_3qdJOR

767, 1.609s JS (bob) sees it. 
775, 1.61s SS sees it.

797, 1.619 bob (js) /keys/query for bob
823, 1.753 alice (rust) SS delivers changed entry for bob
831, 1.781 alice (rust) /keys/query for bob

875, 2.16s bob NEW_DEVICE /login response String value: syt_dXNlci0yLWJvYg_sXHbnXvELCuGoDdoQjAz_0GMdjf
908, 2.183s bob /keys/upload response syt_dXNlci0yLWJvYg_sXHbnXvELCuGoDdoQjAz_0GMdjf

917, 2.184s SS sees bob changed
928, 2.184s bob (JS) sees bob changed

953, 2.193s bob NEW_DEVICE syt_dXNlci0yLWJvYg_sXHbnXvELCuGoDdoQjAz_0GMdjf /keys/query for bob
969, 2.195s bob syt_dXNlci0yLWJvYg_BCEPDhFaFLDZGIrkSuVX_1q3cQp /keys/query for bob
```

Upon close inspection, this seems to be caused because device list updates _don't wake up connections_. This means it takes 30s for the device list update to make it to Alice, and in the meantime messages sent aren't encrypted for Bob's new device. Adding a sleep of at least 30s between Bob's new device login and Alice sending the message fixes the test.

Separately, the code for `func (t *DeviceDataTable) Select(userID, deviceID string, swap bool)` is not atomic, because the transaction lacks `... FOR UPDATE` in the `SELECT`, meaning it is possible for the following to happen:
 - Reader: BEGIN TXN
 - Reader: SELECT device data 
 - Poller gets a device list change entry
 - Writer: BEGIN TXN
 - Writer: SELECT device data
 - Writer: INSERT .. ON CONFLICT DO UPDATE .. [row level lock acquired]
 - Reader: UPDATE ... [blocks due to Writer txn]
 - Writer: COMMIT
 - Reader: UPDATE unblocks
 - Reader: COMMIT
 - Device list change entry was dropped. 

With SELECT .. FOR UPDATE, the lock is acquired at SELECT time meaning this cannot happen and instead:
 - Reader: BEGIN TXN
 - Reader: SELECT device data  [row level lock acquired]
 - Poller gets a device list change entry
 - Writer: BEGIN TXN
 - Writer: SELECT device data [blocks due to Reader txn]
 - Reader: UPDATE
 - Reader: COMMIT
 - Writer: SELECT device data unblocks
 - Writer: INSERT .. ON CONFLICT DO UPDATE .. [row level lock acquired]
 - Writer: COMMIT

If this happens, senders may not have up-to-date device lists for users, causing UTDs for the newly logged in device. For this to happen, the device must login at the same time as the client is sending SS requests.

kegsay:
Fixes #430

This does two things:
 - Fix a hypothetical race condition which could cause a lost update _at insert time_ (due to lack of `.. FOR UPDATE`)
 - Fix an issue which will cause lost device list updates if the response containing the update is A) lost and B) the client starts a new connection for whatever reason e.g timed out, full buffer, server restart.

This adds a regression test and refactors the unit tests to be testing the correct semantics.

History on this is:
 - the original impl worked but was confusing
 - dmr refactored it in https://github.com/matrix-org/sliding-sync/commit/740c14cabcf793e26d9a9b5ab51b31b258760364 but did introduce a bug. The `result` previously had the `New` items put into `Sent` (the original intent) but now we were copying `result` to `writeBack` and _then_ swapping_ `New` to `Sent`. As the caller only looks in `Sent`, it won't see any device list updates. This would fix itself on the _next request_ because we then return `Sent` confusingly. However, the semantics are like this to guard against network failures, and since this refactor specific network failures would drop the update.
 - tests got written using the buggy code as a guide https://github.com/matrix-org/sliding-sync/commit/1fe920d1b68485b7e71111b0d64bfb1b25b9ff23

I found this out when trying to fix [this test](https://github.com/matrix-org/complement-crypto/actions/runs/8999628143/job/24722166536?pr=50#step:15:3006) in complement crypto, which is sensitive to how quickly the device list update will be delivered to the EX client. Hence me finding the bug and fixing it.

This is effectively a UTD cause for the purposes of https://github.com/element-hq/element-meta/issues/245 because it meant EX clients may miss a device list update, causing EX clients to fail to encrypt for newly logged in devices (just like that failing complement crypto test).

### DIFF
--- a/state/device_data_table.go
+++ b/state/device_data_table.go
@@ -70,6 +70,9 @@ func (t *DeviceDataTable) Select(userID, deviceID string, swap bool) (result *in
 		if !swap {
 			return nil // don't swap
 		}
+		// the caller will only look at sent, so make sure what is new is now in sent
+		result.DeviceLists.Sent = result.DeviceLists.New
+
 		// swap over the fields
 		writeBack := *result
 		writeBack.DeviceLists.Sent = result.DeviceLists.New
--- a/state/device_data_table.go
+++ b/state/device_data_table.go
@@ -46,7 +46,7 @@ func NewDeviceDataTable(db *sqlx.DB) *DeviceDataTable {
 func (t *DeviceDataTable) Select(userID, deviceID string, swap bool) (result *internal.DeviceData, err error) {
 	err = sqlutil.WithTransaction(t.db, func(txn *sqlx.Tx) error {
 		var row DeviceDataRow
-		err = txn.Get(&row, `SELECT data FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2`, userID, deviceID)
+		err = txn.Get(&row, `SELECT data FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2 FOR UPDATE`, userID, deviceID)
 		if err != nil {
 			if err == sql.ErrNoRows {
 				// if there is no device data for this user, it's not an error.
@@ -104,7 +104,7 @@ func (t *DeviceDataTable) Upsert(dd *internal.DeviceData) (err error) {
 	err = sqlutil.WithTransaction(t.db, func(txn *sqlx.Tx) error {
 		// select what already exists
 		var row DeviceDataRow
-		err = txn.Get(&row, `SELECT data FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2`, dd.UserID, dd.DeviceID)
+		err = txn.Get(&row, `SELECT data FROM syncv3_device_data WHERE user_id=$1 AND device_id=$2 FOR UPDATE`, dd.UserID, dd.DeviceID)
 		if err != nil && err != sql.ErrNoRows {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


devonh (approved):
Everything checks out to me. The new test logic is clear and easy to follow.

Issue #435, opened 2024-05-16
perf: refactor device data table


kegsay:
The device data table doesn't play well with the MVCC nature of postgres because that table has relatively few but huge rows. Each update is therefore expensive, as it contains a copy of the old state. As the rows aren't being touched frequently enough, it doesn't tend to get detected by the AUTOVACUUM daemon, causing excessive disk usage. This is exacerbated by the nature of the data: a 1:N mapping per unit work (one device logging in results in N updates).

This should be refactored into a new `syncv3_device_list_updates` table which would look something like:
```
user_id | device_id | target_user_id | target_state | bucket
--------+-----------+----------------+--------------+--------
alice   |  FOOBAR   |     bob        |    changed   |    new
alice   |  FOOBAR   |     charlie    |    left      |    new
alice   |  FOOBAR   |     bob        |    left      |    sent

UNIQUE(user_id, device_id, target_user_id, bucket)

new/sent and changed/left will likely be enums.
```
This says, for the client alice|FOOBAR:
 - We sent bob=left and have not got an ACK yet for it
 - We have new changes bob=changed, charlie=left and have not sent it to the client yet.

When the poller gets even more device list updates, it just UPSERTs the value e.g `doris, changed` with bucket=new.

When the client requests device lists, we:
 - Atomically... `begin txn`
 - Delete the now acknowledged changes `delete where user_id=alice, device_id=FOOBAR, bucket=sent`
 - Grab any new updates `select target_user_id, target_state where user_id=alice, device_id=FOOBAR, bucket=new`
 - Mark those updates as sent `update set bucket=sent where user_id=alice, device_id=FOOBAR, bucket=new`
 - `COMMIT`

This means a device list change only results in a single UPSERT, whereas today it involves a deserialisation/serialisation step which is expensive.

This should:
 - reduce poller CPU usage
 - reduce DB load
 - reduce the size of dead tuples
 - increase churn on this new table: however we expect postgres to cope much better with frequently inserted/updated/deleted small rows.

Critically, for the pathological scenario of a client who has gone away never to return, their cost on the DB is much smaller as we are not accumulating an ever growing blob of CBOR.

devonh:
This makes sense to me as an optimization for device list updates.
What will happen with the other device data fields such as one-time keys & fallback keys?
Will those still be in the device data table as a CBOR blob, and just the device list info will be extracted out into it's own table?

S7evinK:
Assuming 
[...]

PR #437, opened 2024-05-17
Refactor device data

kegsay:
Fixes https://github.com/matrix-org/sliding-sync/issues/435

Migration for 10k devices takes around 10-15mins, depending on the size of the accumulated device list updates (particularly monstrous on matrix.org). This isn't great, but isn't awful.


state/device_list_table.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	// grab any 'new' updates
+	result, err = t.selectDeviceListChangesInBucket(txn, userID, deviceID, BucketNew)
+	if err != nil {
+		return nil, err
+	}
+
+	// mark these 'new' updates as 'sent'
+	_, err = txn.Exec(`UPDATE syncv3_device_list_updates SET bucket=$1 WHERE user_id=$2 AND device_id=$3 AND bucket=$4`, BucketSent, userID, deviceID, BucketNew)
+	return result, err
```
devonh, line 120: As you previously mentioned, changing this to use `UPDATE ... RETURNING` might be a better approach.

state/migrations/20240517104423_device_list_table_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = upDeviceListTable(ctx, tx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tx.Commit()
+
```
devonh, line 73: This test is missing the bit that actually ensures we didn't lose any data during the migration.

state/migrations/20240517104423_device_list_table.go
```diff
@@ -0,0 +1,188 @@
...
+		_, err = tx.ExecContext(ctx, `UPDATE syncv3_device_data SET data=$1 WHERE user_id=$2 AND device_id=$3`, data, userID, deviceID)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func downDeviceListTable(ctx context.Context, tx *sql.Tx) error {
+	// no-op: we'll drop the device list updates but still work correctly as new/sent are still in the cbor but are empty
```
devonh, line 171: Shouldn't the down migration still be done for completeness?
Or is the theory here that future syncs will rebuild the device list updates in the cbor well enough that we don't really care to do the down migration?

Also - does this need to drop the `syncv3_device_list_updates` table? Or does that happen elsewhere?

devonh (changes requested):
Overall this looks great!

PR #439, opened 2024-05-20
fix: urlencode since opaque string

gnieto:
Since/next batch is an opaque string and might need to be urlencoded before being sent to the server.

### Pull Request Checklist


kegsay (approved):
Good spot, thanks.

Issue #415, opened 2024-04-04
Pass secret as file path for security


V02460:
The sliding sync proxy currently only takes its secret via the environment variable SYNCV3_SECRET. When used with systemd this is not considered secure:

> Note that environment variables are not suitable for passing secrets (such as passwords, key material, …) to service processes. Environment variables set for a unit are exposed to unprivileged clients via D-Bus IPC, and generally not understood as being data that requires protection. Moreover, environment variables are propagated down the process tree, including across security boundaries (such as setuid/setgid executables), and hence might leak to processes that should not have access to the secret data.

From https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment

Please change the sliding sync proxy to accept the path to a secret file.

pcolladosoto:
Hi @V02460! I'd be more than happy to take a stab at this. I could have something working towards the end of the week: does that sound good?

pcolladosoto:
This will hopefully be introduced on PR #446!

Zelaf:
For more people coming across this and wish to have more security, I've come up with a clever workaround!

I created a script I simply named `run.sh`:
```sh
#!/bin/sh

## Get environmental variables

# PostgreSQL authentication
psql_username=$(cat /config/secrets/postgresql_username)
psql_password=$(cat /config/secrets/postgresql_password)
export SYNCV3_DB="postgres://${psql_username}:${psql_password}@postgresql:5432/syncserver?sslmode=disable"

# Matrix sync proxy secret
export SYNCV3_SECRET=$(cat /config/secrets/.secret)

# Matrix sync proxy server name
export SYNCV3_SERVER="https://your.homeserver.here"

# Setting bind address
export SYNCV3_BINDADDR=":8009"

## Run sync server
env /usr/bin/syncv3
```

Then in the same directory I created a directory called `secrets` where I keep three files for the authentication.
```
sliding-sync/
|-- run.sh
|-- secrets/
|   |-- .secret
|   |-- postgresql_username
|   |-- postgresql_password
```
`postgresql_username`: Username of the PostgreSQL server
`postgresql_password`: Password for the user on the PostgreSQL server
`.secret`: The generated secret file.

Then for the docker compose I did:
```yaml
  sliding_sync:
    image: ghcr.io/matrix-org/sliding-sync:latest
    ports:
      - 8009:8009
    volumes:
      - ./sliding-sync:/config
[...]

PR #446, opened 2024-05-30
Add support for a JSON configuration file

pcolladosoto:
This PR strives to add support for specifying the configuration options through a file. This PR is motivated by the discussion on #415.

In order to keep changes as smooth as possible we have chosen to leverage the name of the existing environment variables as the JSON keys. We have also chosen JSON so as to avoid delving into parsing logic; besides, JSON encoding/decoding are part of the standard library.

The path to the configuration file is to be passed through the `-conf` flag.

In order to remain backwards-compatible, environment variables will always override settings specified on the JSON file. The default values for settings will also be honoured.

At any rate, we're more than happy to change anything that might be needed.

Thanks a ton for your time!

### Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ SYNCV3_LOG_LEVEL     Default: info. The level of verbosity for messages logged. 
 SYNCV3_MAX_DB_CONN   Default: unset. Max database connections to use when communicating with postgres. Unset or 0 means no limit.
 ```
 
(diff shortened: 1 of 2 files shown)


pcolladosoto:
Hi @kegsay! I'm pinging you because the last commit on the repo is yours... Is there any intention of moving froward with this PR? If not I'd be more than happy to close it to try and clean up the clutter...

Thanks for your time and sorry for the noise...

PR #449, opened 2024-06-04
Protect against dropped databases

kegsay:


devonh (changes requested):
Changes look good. The comment is much appreciated, especially for future people trying to understand this code.

Would it be simple enough to add a test for this? I imagine something around here would be easiest: https://github.com/matrix-org/sliding-sync/blob/5da6865609a32ae07940b383e6baed78523ad343/tests-integration/extensions_test.go#L273 All that should be necessary is to start up a fresh test, add some to-device messages, and do a sync with a position > 0. Then check you received all the messages. That should mimic the DB dropping sequence, without having to go through actually doing that in a test.

tests-integration/extensions_test.go
```diff
@@ -443,6 +444,53 @@ func TestExtensionToDevice(t *testing.T) {
...
+				Core:  extensions.Core{Enabled: &boolTrue},
+				Since: hiSince,
+			},
+		},
+	})
+	m.MatchResponse(t, res, m.MatchList("a", m.MatchV3Count(0)), m.MatchToDeviceMessages(toDeviceMsgs), func(res *sync3.Response) error {
+		// ensure that we return a lower numbered since token
+		if res.Extensions.ToDevice.NextBatch == hiSince {
+			return fmt.Errorf("next_batch got %v wanted lower", hiSince)
+		}
```
devonh, line 489: I agree with the comment here.
But the check isn't actually checking for a lower number.

devonh:
New test looks good! (other than the token comparison)

devonh (approved):

